PR #149, opened 2023-11-20
Changes to enable fp8 on multi devices

y-sq:
- If the model is casted to bf16 (model = model.to(get_torch_dtype(dtype))), dtype = bf16 is also passed to the scale_a parameter of the float8 tensor, and caused
```
output, output_amax = torch._scaled_mm(
RuntimeError: scale_a must be float scalar
```
- The float8Linear classes used in multi-gpu are Float8Column/RowParallelLinear, sync_float8_amax_and_scale_history needs to identify these class types.

### DIFF
--- a/float8_experimental/float8_linear_utils.py
+++ b/float8_experimental/float8_linear_utils.py
@@ -143,10 +143,10 @@ def sync_float8_amax_and_scale_history(
     if fp8_classes is None:
         from float8_experimental.float8_linear import Float8Linear
 
-        fp8_classes = [Float8Linear]
+        fp8_classes = Float8Linear
 
     for name, child in model.named_modules():
-        if not any(isinstance(child, a) for a in fp8_classes):
+        if not isinstance(child, fp8_classes):
             continue
 
         #
--- a/float8_experimental/float8_linear_utils.py
+++ b/float8_experimental/float8_linear_utils.py
@@ -142,6 +142,7 @@ def sync_float8_amax_and_scale_history(
 
     if fp8_classes is None:
         from float8_experimental.float8_linear import Float8Linear
+
         fp8_classes = [Float8Linear]
 
     for name, child in model.named_modules():
--- a/float8_experimental/float8_linear_utils.py
+++ b/float8_experimental/float8_linear_utils.py
@@ -117,7 +117,9 @@ def swap_linear_with_float8_linear(
             swap_linear_with_float8_linear(child, module, emulate)
 
 
-def sync_float8_amax_and_scale_history(model: torch.nn.Module) -> None:
+def sync_float8_amax_and_scale_history(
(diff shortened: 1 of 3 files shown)


float8_experimental/float8_linear_utils.py
```diff
@@ -138,10 +140,13 @@ def sync_float8_amax_and_scale_history(model: torch.nn.Module) -> None:
...
 
-    from float8_experimental.float8_linear import Float8Linear
+    if fp8_classes is None:
+        from float8_experimental.float8_linear import Float8Linear
+
+        fp8_classes = [Float8Linear]
 
     for name, child in model.named_modules():
-        if not isinstance(child, (Float8Linear)):
+        if not any(isinstance(child, a) for a in fp8_classes):
```
drisspg, line 149: Since we have removed the NoTs class I think this is likely rebase buggies

drisspg:
Looks great, can we add a test?

float8_experimental/float8_linear_utils.py
y-sq: In multi-gpu cases, we have Float8ColumnParallelLinear and Float8RowParallelLinear (which have dependencies of external distributed training code) as the fp8 classes. So I modified here to pass the class types to sync_float8_amax_and_scale_history.

y-sq:
It's strange that format check passed on my side but failed on github checks..
```
ufmt --version
ufmt, version 2.3.0
ufmt check .
✨ 22 files already formatted ✨
```

drisspg:
Dont worry about the format, there is deviation between ufmt and internal, and I can't pin it down, so those were not likely caused by anything you did

test/test_base.py
```diff
@@ -177,6 +177,45 @@ def test_linear_float8_weight_tag(self):
...
+        m_ref = nn.Linear(16, 32, bias=True, device="cuda", dtype=linear_dtype)
+        self._test_linear_impl(x, m_ref, linear_type, emulate)
+
+        m = nn.Linear(32, 16, device="cuda", dtype=linear_dtype)
+        m = Float8Linear.from_float(m, emulate)
+
+        # Cast the module to dtype
+        m = m.to(dtype=linear_dtype)
+
+        # autocast off
```
drisspg, line 200: Nit: this test does cover it but also could we assert the buffer types are still fp32

drisspg (approved):

float8_experimental/float8_linear_utils.py
drisspg: ahhh I see I was misreading this, but if we make fp8_classes a Tuple[types] couldn't we still keep the check as is, this is a nit anyways both accomplish the same thing

drisspg (approved):
Thanks 😄 !

PR #155, opened 2023-12-15
Test fsdp compile

y-sq:
- Added an initial benchmark for fsdp + fp8, benchmarks/bench_multi_gpu.py
  - Note that the benchmark has issues with "compile", please see the experiments doc for more context.
  - CUDA_VISIBLE_DEVICES=0,1 python benchmarks/bench_multi_gpu.py 2>&1 | tee logs

- Modified the existing test/test_fsdp.py to include compile.
  - The script can success with compile, but it's with a workaround. Please see the comment there for context.
  - Added relu layers after linear layers so that the numerics can be stable after several iterations and can be used to check accuracy results.

- In bench/profile_linear_float8.py, changed the TE FP8 receipt to be the same as PyTorch FP8.

- In sync_float8_amax_and_scale_history, added an option to combine all_reduce calls into one call.
  - Need to verify if this can bring performance gain.
  - Another option is to let each layer handle its own amax update and sync, which is the approach used by TE fp8.

### DIFF
--- a/benchmarks/bench_multi_gpu.py
+++ b/benchmarks/bench_multi_gpu.py
@@ -1,0 +1,247 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+Test numerics of single GPU vs FSDP of toy model. At a high level:
+1. start with reference input and state dict for a single GPU model
+2. run fw+bw+optim on single GPU, save the results
+3. run fw+bw+optim with FSDP, save the results
+4. verify that the outputs and state dicts after optim update match
+
+later 1-4 can be repeated for fp16, various combinations of fp8, etc.
+"""
+
+import os
+import warnings
+from typing import Callable, List, Optional, Tuple
+
+import fire
+
+import torch
+import torch.distributed as dist
+import torch.multiprocessing as mp
+import torch.nn as nn
+import torch.utils.benchmark as benchmark
+from float8_experimental.float8_linear import Float8Linear
+from float8_experimental.float8_linear_utils import (
+    swap_linear_with_float8_linear,
+    sync_float8_amax_and_scale_history,
+    update_float8_amax_and_scale_history,
+)
+from torch.distributed.fsdp import (
+    FullStateDictConfig,
+    FullyShardedDataParallel as FSDP,
+    StateDictType,
+)
+
+# Check if transformer_engine is installed
+transformer_engine_installed = False
+try:
+    import transformer_engine.pytorch as te
+    from transformer_engine.common import recipe
+
+    transformer_engine_installed = True
+except ImportError:
+    print("transformer_engine not installed and we won't compare against this")
+
+
+torch.manual_seed(0)
+
+B, M, K, N = 32, 32, 32, 32
+lr = 0.01
+N_ITER = 11
+
+def benchmark_torch_function_in_microseconds(
+    func: Callable,
+    *args,
+    **kwargs,
+) -> float:
+    t0 = benchmark.Timer(
+        stmt="func(*args, **kwargs)",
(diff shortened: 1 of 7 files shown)


y-sq:
For the failed check, I verified "ufmt check ." locally which gave no error.

test/test_fsdp.py
```diff
@@ -83,7 +86,7 @@ def fsdp_main(rank, world_size, args):
         rank
     )
     model.load_state_dict(torch.load(sd_in_fname))
-    model = FSDP(model)
+    model = FSDP(model, use_orig_params=COMPILE)
```
vkuzo, line 89: just set use_orig_params to True, and add a comment that compile only supports this setting?

test/test_fsdp.py
```diff
@@ -97,12 +100,21 @@ def fsdp_main(rank, world_size, args):
...
         y_local.sum().backward()
         sync_float8_amax_and_scale_history(model)
         optimizer.step()
+        return y_local
+
+    # TODO: Fix here. Running for one iteration and then compile is a workaround for compiling fsdp float8.
+    # A better way for compile is to FSDP(torch.compile(model)).
+    forward_backward()
+    if COMPILE:
+        forward_backward = torch.compile(forward_backward)
```
vkuzo, line 115: are we sure this is supported?

how about something simpler such as

```
model = torch.compile(model)
sync_func = torch.compile(sync_float8_amax_and_scale_history)
for _ in range(N_ITER):
   ... loop as before ...
```

float8_experimental/float8_linear_utils.py
```diff
@@ -117,8 +117,23 @@ def swap_linear_with_float8_linear(
...
+    fp8_layers = [
+        child for name, child in model.named_modules() if isinstance(child, fp8_classes)
+    ]
+
+    return fp8_layers
+
+
 def sync_float8_amax_and_scale_history(
-    model: torch.nn.Module, fp8_classes=None
+    model: torch.nn.Module, fp8_classes=None, fp8_layers=None, combine_reduction=False
```
vkuzo, line 136: thoughts on splitting these changes into a separate diff? something like
1. diff 1: add microbenchmark for FSDP
2. diffs 2..n: improvements to sync function, with before+after benchmark numbers

benchmarks/bench_multi_gpu.py
```diff
@@ -0,0 +1,241 @@
...
+
+def run():
+    compile = True
+    base_dtype = torch.bfloat16
+    print(f"{base_dtype = }")
+    print(f"{compile = }")
+    # generate input data
+    ref_input = torch.randn(B, M, K).cuda().to(base_dtype)
+    # run fsdp model
+    WORLD_SIZE = torch.cuda.device_count()
```
vkuzo, line 234: maybe make this configurable? usually easier to debug issues and iterate with world_size == 2, and then verify the fix with world_size == torch.cuda.device_count()

benchmarks/bench_multi_gpu.py
```diff
@@ -0,0 +1,241 @@
...
+    print(f"{WORLD_SIZE = }")
+    args = (base_dtype, ref_input, compile)
+    mp.spawn(fsdp_main, args=(WORLD_SIZE, args), nprocs=WORLD_SIZE, join=True)
+
+
+if __name__ == "__main__":
```
vkuzo, line 240: does this need a shell script wrapper like the other distributed benchmarks in this file?  if not, can you add a comment with example way to call this?

float8_experimental/float8_linear_utils.py
y-sq: Thank you for the comments. Let me split into small PRs.

test/test_fsdp.py
y-sq: This did work in fsdp.py test. But let me change it to the officially supported `FSDP(torch.compile(model))`.

y-sq:
Will split and close this one.

PR #157, opened 2023-12-15
add LLaMa 7b e2e fw+bw benchmark and profiler

vkuzo:
Summary:

This is just taking @drisspg's
https://github.com/drisspg/lit-gpt/blob/fp8_train/pretrain/fp8_openweb.py script and packaging a lightweight version of it in this repo.

A couple of notes:
1. we do some hacks to import lit-gpt to avoid taking it as a dependency of this repo, we can make this more robust later if needed
2. all data loading is mocked out
3. we add a `skip_n_encoder_layers` argument to make it easier to iterate on traces and inspect inductor code, and get incremental results faster

Test Plan:

```
python benchmarks/bench_encoder_block.py --skip_n_encoder_blocks=31 --fp8_linear_type delayed --profile True --profile_path ../tmp/test.json --compile True
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/benchmarks/bench_encoder_block.py
+++ b/benchmarks/bench_encoder_block.py
@@ -1,0 +1,240 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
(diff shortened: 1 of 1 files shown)


vkuzo:
added it here instead: https://github.com/pytorch-labs/ao_benchmarks/pull/155

PR #163, opened 2023-12-15
Combine amax reduction calls

y-sq:
~~Add an option to combine the amax sync reduction~~ (Use combine-reduction as the default behavior)
- Combine the reduction call of each type amax scaling factor (totally 3 all_reduce calls). We can also further combine them into one single call.
  - Verified other tests can still pass. So we don't need to change existing benchmark code.
    - pytest test/test_base.py
    - ./test/test_fsdp.sh 
- Tested the new option using small llama models with 8 fsdp groups. Time taken by sync_float8_amax_and_scale_history reduced from **29ms**[1] to **10ms**[2].

[1] Traces without combine reduction, https://www.internalfb.com/intern/perfdoctor/trace_view?filepath=tree/traces/trace.138932292910521.json.gz&bucket=acadia
[2] https://www.internalfb.com/intern/perfdoctor/trace_view?filepath=tree/traces/trace.202842416426594.json.gz&bucket=acadia
\* Results from trace[2] was updated to the correct number.
\*\* Need Meta internal access to open these traces.

### DIFF
--- a/float8_experimental/float8_linear_utils.py
+++ b/float8_experimental/float8_linear_utils.py
@@ -87,8 +87,20 @@ def swap_linear_with_float8_linear(
             swap_linear_with_float8_linear(child, module, emulate)
 
 
+def get_float8_layers(model: torch.nn.Module, fp8_classes=None):
+    if fp8_classes is None:
+        fp8_classes = Float8Linear
+
+    # Get all fp8 layers and tensors
+    fp8_layers = [
+        child for name, child in model.named_modules() if isinstance(child, fp8_classes)
+    ]
+
+    return fp8_layers
+
+
 def sync_float8_amax_and_scale_history(
-    model: torch.nn.Module, fp8_classes=None
+    model: torch.nn.Module, fp8_classes=None, fp8_layers=None
 ) -> None:
     """
     Manages the float8 amax and scale bookkeeping. In detail, it does the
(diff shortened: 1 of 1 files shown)


float8_experimental/float8_linear_utils.py
```diff
@@ -138,24 +153,50 @@ def sync_float8_amax_and_scale_history(
...
+                device="cuda",
+                requires_grad=False,
+            )
+            fp8_amax_dL_dY_tensor = torch.tensor(
+                [child.fp8_amax_dL_dY for child in fp8_layers],
+                dtype=torch.float32,
+                device="cuda",
+                requires_grad=False,
+            )
+            # print("fp8_amax_x_tensor, ", fp8_amax_x_tensor)
```
drisspg, line 181: nit: probs remove right?

float8_experimental/float8_linear_utils.py
```diff
@@ -117,8 +117,23 @@ def swap_linear_with_float8_linear(
...
+    # Get all fp8 layers and tensors
+    fp8_layers = [
+        child for name, child in model.named_modules() if isinstance(child, fp8_classes)
+    ]
+
+    return fp8_layers
+
+
 def sync_float8_amax_and_scale_history(
-    model: torch.nn.Module, fp8_classes=None
```
drisspg, line 121: we should probs document this function and also what the new args do / how they should be used, since I assume you get all the fp8_layers once and then pass that in every iteration

drisspg (approved):
Some small comments and I think you need to run `ufmt format .` but otherwise awesome speed ups!!

float8_experimental/float8_linear_utils.py
```diff
@@ -138,24 +153,50 @@ def sync_float8_amax_and_scale_history(
...
+            )
+            fp8_amax_dL_dY_tensor = torch.tensor(
+                [child.fp8_amax_dL_dY for child in fp8_layers],
+                dtype=torch.float32,
+                device="cuda",
+                requires_grad=False,
+            )
+            # print("fp8_amax_x_tensor, ", fp8_amax_x_tensor)
+            dist.all_reduce(fp8_amax_x_tensor, op=dist.ReduceOp.MAX)
+            dist.all_reduce(fp8_amax_x_tensor, op=dist.ReduceOp.MAX)
```
vkuzo, line 183: fp8_amax_w_tensor?

float8_experimental/float8_linear_utils.py
```diff
@@ -138,24 +153,50 @@ def sync_float8_amax_and_scale_history(
...
+            dist.all_reduce(fp8_amax_x_tensor, op=dist.ReduceOp.MAX)
```
vkuzo, line 184: dL_dY?

float8_experimental/float8_linear_utils.py
drisspg: ohh damn good catch

float8_experimental/float8_linear_utils.py
```diff
@@ -117,8 +117,23 @@ def swap_linear_with_float8_linear(
...
+    fp8_layers = [
+        child for name, child in model.named_modules() if isinstance(child, fp8_classes)
+    ]
+
+    return fp8_layers
+
+
 def sync_float8_amax_and_scale_history(
-    model: torch.nn.Module, fp8_classes=None
+    model: torch.nn.Module, fp8_classes=None, fp8_layers=None, combine_reduction=False
```
vkuzo, line 136: can we measure time on single GPU, and if that's a net positive as well just delete the old code path?  It would be great to keep things simple.

y-sq: Did you mean the `combine_reduction` option or the `fp8_layers` option? 

For `combine_reduction`, I think we can remove it and keep `combine_reduction=True` as default. 

For `fp_layers`, we have many exiting tests and benchmarks (such as single-gpu llama_7B benchmarks) that using the original call (`sync_float8_amax_and_scale_history(model)`). So I kept it as optional and can support `None`

float8_experimental/float8_linear_utils.py
y-sq: Thanks!!!

y-sq:
Updates:
- Add comments and remove unused comments
- Remove the option of "combine_reduction", always combine reduction instead
- Fix to use the correct scaling factors of fp8_amax_w_tensor and fp8_amax_dL_dY_tensor

y-sq:
I ran the format check on my devgpu server, which didn't give any errors:
```
$ ufmt check .
✨ 22 files already formatted ✨
```
However, the check on github still failed.

PR #165, opened 2023-12-18
Move testing dependencies to optional dependency

drisspg:
# Summary
Moved depdencies that are only used for tests or for benchmarking to optional depdency

Also fix format

Usage would be 
`pip install - e .[test]`


vkuzo (approved):
maybe update the readme too?

PR #170, opened 2023-12-26
[wip] hooks

vkuzo:
Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


test/test_bw_hook.py
```diff
@@ -0,0 +1,32 @@
+import torch
+import torch.nn as nn
+
+class TestAutogradFunction(torch.autograd.Function):
```
vkuzo, line 4: this example works as expected, fwiw

float8_experimental/dynamic_linear/dynamic_float8_linear.py
```diff
@@ -38,6 +39,24 @@ def backward(ctx, gradY):
             None,
         )
 
+def cast_x_to_float8_e4m3fn_pre_hook(module, args):
+    """
+    Hook to cast the incoming activation to `torch.float8_e4m3fn`
+    """
+    return module.cast_to_float8(args[0])
+
+def cast_dldy_to_float8_e5m2_pre_hook(module, grad_output):
```
vkuzo, line 48: what I do in this function doesn't seem to affect the gradient received by `torch.mm`, unless I modify `grad_output` inplace

float8_experimental/dynamic_linear/dynamic_float8_linear.py
```diff
@@ -38,6 +39,20 @@ def backward(ctx, gradY):
...
+    Hook to cast the incoming activation to `torch.float8_e4m3fn`
+    """
+    return module.cast_to_float8(args[0])
+
+
+def cast_dldy_to_float8_e5m2_forward_hook(module, args, output):
+    """
+    Hook to cast the incoming gradient to `torch.float8_e5m2`
+    """
+    new_output = NoopFwToFloat8E5M2Bw.apply(output, module.emulate)
```
wanchaol, line 53: I'm a bit surprised that torch.compile does not support this case with autograd function, as both DTensor pre-forward hook and forward hook are using autograd functions, probably because DTensor is using sth like `allow_in_graph` in dynamo. Maybe one way to workaround is to use a function that calls `NoopFwToFloat8E5M2Bw.apply(output, module.emulate)` inside, and then make it a allow_in_graph in dynamo?

vkuzo:
note: backward pre hook works with https://github.com/pytorch/pytorch/pull/116454/

vkuzo:
@wanchaol , looks like the current dynamo issue is lack of support of `Float8Tensor` being in a subgraph boundary.  Logs with evidence supporting this: https://gist.github.com/vkuzo/3cf42b6a54a68be1a45e8e1ca07eeb26.  https://github.com/pytorch-labs/float8_experimental/pull/166 has a more minimal repro.

Would you expect `allow_in_graph` to be related?

PR #172, opened 2023-12-31
enable autocast + compile + FSDP + Float8Linear

vkuzo:
Summary:

This adds a couple of config options to unbreak autocast + compile + FSDP + Float8Linear. To enable these options, the user needs to do:

```
config.enable_amax_init = False
config.enable_pre_and_post_forward = False
```

The `enable_amax_init` config adds the option to disable amax initialization. The reason this is currently broken is:
1. FSDP is not full-graph friendly (regardless of compile)
2. the amax init function has a graph break in distributed code because it uses inplace distributed collectives.  I did try to use functional collectives (https://github.com/pytorch-labs/float8_experimental/pull/171), but that ran into numerical issues with compile, so for now just working around it.
3. graph breaks in Float8Linear code are not supported because of the issue documented in https://github.com/pytorch-labs/float8_experimental/pull/166
4. so, as a workaround for all of the above, we just skip amax init for now.  We do know from NVIDIA that this path is not needed for model convergence, and TE does not support this at all. It was nice for testing but not necessary for training jobs.

The second config option disables pre-forward and post-forward. I don't have a repro in a unit test for now, but this does unbreak LLaMa 7B on 8 GPUs with FSDP + compile. Specifically, the thing which is broken in pre-forward/post-forward is assignment on module attributes. My hunch is that this graph breaks if autocast + FSDP are on, and graph breaks are not supported due to (3) above.

Test Plan:

```
// unit / integration tests
with-proxy test/test_everything.sh

// run the LLaMa 7b trainer on 8 GPUs with autocast + compile + FSDP + Float8Linear, no compile errors
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/float8_experimental/float8_linear.py
+++ b/float8_experimental/float8_linear.py
@@ -156,14 +156,14 @@ class Float8LinearMixin(object):
         # Note: is_amax_initialized is not a buffer to avoid data dependent
         # control flow visible to dynamo
         # TODO(future PR): add serialization for this flag
-        self.is_amax_initialized = False
+        self.is_amax_initialized = not config.enable_amax_init
 
         # Syncing of amaxes and scales happens outside of this function. This
         # flag is here to enforce that the user does not forget to do this.
-        self.amax_and_scale_synced = False
+        self.amax_and_scale_synced = not config.enable_amax_init
 
         # This is needed to properly handle autocast in the amax/scale
-        # update function
+        # update function for torch.float16
         self.last_seen_input_dtype = None
 
         # If true, this enables TP+SP style distributed comms in TP primitives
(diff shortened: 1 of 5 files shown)


float8_experimental/float8_linear.py
```diff
@@ -335,7 +342,7 @@ def forward(self, x):
         y = self.cast_y_to_float8_in_bw(y, self.emulate)
 
         if self.bias is not None:
-            y = y + self.bias.to(self.bias_dtype)
+            y = y + self.bias.to(y.dtype)
```
vkuzo, line 345: this change is not called out in the configs, but this just removes the need to store a module attribute, which also makes this code more full-graph compile friendly.   Modifying module attributes seems to graph break if autocast and FSDP are both enabled.

drisspg (approved):
Okay this makes sense

drisspg:
formatting

PR #173, opened 2024-01-02
[wip] makes dynamic linear support weight caching

vkuzo:
Summary:

Works on single GPU, but not yet with FSDP

FSDP error (from LLaMa trainer), looks like weight_gradient is off by a factor of world_size:
https://gist.github.com/vkuzo/a6fd86085ea1fb05618dbd10fb7f00d1

Test Plan:

```
pytest test/test_base.py -s --sw -k weight_caching

// TODO before land: add test with autocast and FSDP
```

Reviewers:

Subscribers:

Tasks:

Tags:


vkuzo:
not useful

PR #174, opened 2024-01-02
[not for land, yet] example of FSDP with dynamic float8 linear

vkuzo:
Summary:

TODO before land write this up and also add to readme, and reuse code better

Test Plan:

```
CUDA_VISIBLE_DEVICES=7,8 python test/test_dynamic_fsdp_compile.py
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/test/test_dynamic_fsdp_compile.py
+++ b/test/test_dynamic_fsdp_compile.py
@@ -1,0 +1,110 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD 3-Clause license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+Smoke test of autocast + torch.compile + FSDP + Float8DynamicLinear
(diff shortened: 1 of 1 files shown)


test/test_dynamic_fsdp_compile.py
```diff
@@ -0,0 +1,110 @@
...
+def cleanup():
+    dist.destroy_process_group()
+
+
+def get_model(K, N, is_fp8, emulate, base_dtype=torch.float32):
+    m = nn.Sequential(
+        nn.Linear(K, N, dtype=base_dtype),
+        nn.ReLU(),
+    )
+    swap_linear_with_float8_linear(m, Float8DynamicLinear, emulate=emulate)
```
vkuzo, line 57: cc @wanchaol  this is the only line we need to use Float8DynamicLinear for the model

vkuzo: with emulate=False

PR #175, opened 2024-01-02
add dynamic linear to __init__

wanchaol:
This is needed otherwise there will be import errors like:

```
    from float8_experimental.dynamic_linear.dynamic_float8_linear import Float8DynamicLinear
ModuleNotFoundError: No module named 'float8_experimental.dynamic_linear'
```


vkuzo (approved):
thank you!

PR #180, opened 2024-01-10
move dynamic linear to main directory

vkuzo:
Summary:

perf and accuracy are great, so makes more sense to have this at the same level of importance as `Float8Linear`

Test Plan:

```
./test/test_everything.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:


drisspg (approved):
Some conflicts but looks good

PR #181, opened 2024-01-10
remove weight caching

vkuzo:
Summary:

Removes most of
https://github.com/pytorch-labs/float8_experimental/pull/164

This isn't useful in the short term since it doesn't compose with FSDP + compile, and memory overhead is high.  We can bring it back later if needed.

Test Plan:

```
./test/test_everything.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:


drisspg (approved):
Looks good

PR #183, opened 2024-01-11
Update README.md

vkuzo:


README.md
```diff
@@ -19,7 +20,32 @@ pip install -e ".[dev]"
 
 # User API, subject to change
 
-## single GPU
+We provide two types of scaling for per-tensor scaling of tensors: dynamic and delayed.
```
drisspg, line 23: This sounds kinda of weird, what "We provide two scaling strategies: per-tensor dynamic and delayed."

README.md
```diff
@@ -28,124 +54,44 @@ from float8_experimental.float8_linear_utils import (
...
-# create fp32 model
+# create model
 m = Model(...)
 
 # convert all `torch.nn.Linear` modules to `Float8Linear`
 swap_linear_with_float8_linear(m, Float8Linear)
 
+# optional: use FSDP. Note that workarounds are needed for autocast+compile+FSDP+float8 to work
+from float8_experimental import config
+config.enable_amax_init = False
```
drisspg, line 65: can we do like

if "fsdp" do the config stuff otherwise you don't need to

README.md
```diff
@@ -28,124 +54,44 @@ from float8_experimental.float8_linear_utils import (
...
-# the rest of the flow is the same as the single GPU flow
-```
-
-# high level technical design
-
-## UX
-
-We are using a module swap UX to keep things simple. If the user model has `torch.nn.Linear` modules or their `fairscale` TP/SP equivalents,
-we can convert them to float8. `F.linear`, `torch.mm`, `torch.matmul` are not supported at the moment.
+# upcoming work
```
drisspg, line 85: Maybe we make a tracking issue for upcoming work and link that here. Just so we don't forget to remove from reademe, either way is okay though

README.md
```diff
@@ -28,124 +54,44 @@ from float8_experimental.float8_linear_utils import (
...
-We are using tensor subclasses (`Float8Tensor`) to write modular code which satisfies
-autograd's restriction that `x.dtype == x.grad.dtype`.  The way we achieve this is by
-ensuring that instances of `Float8Tensor` set their dtype attribute to the original
-dtype (float32/float16/bfloat16) while the underlying data representation is in float8.
-If you look in `float8_linear.py` and `te_linear.py`, you will see that we pass instances of `Float8Tensor`
-around various `torch.autograd.Function` calls, enabling us to have modular code.
-
-## multi GPU
-
-### TP/SP
```
drisspg, line 103: Maybe we still have the TP/SP stuff documented just saying we know this doesn't work with compile and planto have dtensor integration for this

README.md
```diff
@@ -28,124 +54,44 @@ from float8_experimental.float8_linear_utils import (
...
-
-### separation of concerns
-
-1. `Float8Linear` owns casting X, W and dL/dY to float8 and does all the bookkeeping of the amax, amax_history and scale buffers
-2. user is responsible for applying `Float8Linear` to the right parts of their model with module swaps
-
-
-### Tensor subclasses
-
-We are using tensor subclasses (`Float8Tensor`) to write modular code which satisfies
```
drisspg, line 94: I think this is still fine no?

README.md
```diff
@@ -28,124 +54,44 @@ from float8_experimental.float8_linear_utils import (
...
-
-# if applicable, enable sequence parallel on the right modules
-# TODO make the API for this nicer
-model.foo.bar.fc1.sequence_parallel = True
-model.foo.bar.fc2.sequence_parallel = True
-
-# the rest of the flow is the same as the single GPU flow
-```
-
-# high level technical design
```
drisspg, line 67: same with this, I don't mind the high level design

README.md
```diff
@@ -28,124 +54,44 @@ from float8_experimental.float8_linear_utils import (
...
-We are working on a new FSDP implementation that uses per-parameter sharding that will allow flexible fp8 all-gather. This is being prototyped currently.
+* enable recomputation of float8 casted activations and weights in the backward, to reduce peak memory usage
+* compose with DTensor to enable TensorParallel/SequenceParallel in a way composable with torch.compile
 
 # code tips
 
 * `float8_experimental/float8_linear.py` - `Float8Linear` (main user facing entry point for delayed scaling)
 * `float8_experimental/float8_dynamic_linear.py` - `Float8DynamicLinear` (main user facing entry point for dynamic scaling)
 * `float8_experimental/float8_tensor.py` - `Float8Tensor`, which allows `Float8Linear` to abide by the `x.dtype == x.grad.dtype` restriction
-* `float8_experimental/tp_linear.py` - `Float8ColumnParallelLinear` / `Float8RowParallelLinear` (TP/SP versions of float8 linear)
```
drisspg, line 148: same with above

drisspg (approved):

README.md
vkuzo: I think all of the below is slightly outdated and also more of dev docs than user README.  I think we can have an issue about high level design, it would be nice to edit that without OSS->Meta PR sync.  I can copy-paste this there.

README.md
vkuzo: hmm, it's not that usable without compile though :(  I'm more excited about just deleting this

README.md
vkuzo: will copy-paste to design

PR #190, opened 2024-01-17
match_prefix for test_fsdp against compiled state dict

drisspg:
# Summary

Fix test/test_fsdp.py to match correct state dict entries


awgu (approved):
Nice!

Will we always compile FSDP at the root module? How are other places handling this `_orig_mod.` prefix?

bdhirsh (approved):

drisspg:
@awgu this is our only test doing the state dict matching, so I think this is the only place where this error comes up

PR #186, opened 2024-01-13
Add option for recomputing the casted weight during backwards

drisspg:
# Summary
See: https://github.com/pytorch-labs/float8_experimental/issues/185
For more detail

PR #193, opened 2024-01-23
Checkpoint to reduce fp8_weight tensor saved for backwards

drisspg:
# Summary

Take two of doing this: https://github.com/pytorch-labs/float8_experimental/pull/186
Issue: https://github.com/pytorch-labs/float8_experimental/issues/185

I have verified that the the transpose_cast in backwards is fusable with inductor, now we just gotta coax it into doing what we want..

cc @bdhirsh


drisspg:
I figure this is worth landing over the other since we anticipate that this will be supported by min-cut-partioner @bdhirsh 

I need to add the same for delayed scaling and ensure that that does indeed checkpoint correctly

PR #198, opened 2024-01-30
Add for option to use tensor hooks for Dynamic Linear

drisspg:
# Summary

This is a duplicate of: https://github.com/pytorch-labs/float8_experimental/pull/170
With more testing, ideally I think we wouldn't have the choice between hooks and modified forwards and just use hooks. However compile does not appear to support this yet

### DIFF
--- a/float8_experimental/config.py
+++ b/float8_experimental/config.py
@@ -14,3 +14,8 @@ enable_amax_init = True
 # this doesn't work with autocast + torch.compile + FSDP. Enabling this
 # option is useful for safety, but not strictly necessary.
 enable_pre_and_post_forward = True
+
+# If True, dynamic linear uses hooks for activation casting
+# TODO(before land): add test coverage for both cases
+# dynamic_use_activation_hooks = True
+# dynamic_use_activation_hooks = False
--- a/float8_experimental/float8_dynamic_linear.py
+++ b/float8_experimental/float8_dynamic_linear.py
@@ -7,9 +7,10 @@
 A wrapper around a `torch.nn.Linear` module which does fp8 compute.
 """
 
+import float8_experimental.config as config
 import torch
 
-from float8_experimental.float8_tensor import Float8Tensor
+from float8_experimental.float8_tensor import Float8Tensor, to_fp8_no_autograd
 from float8_experimental.float8_utils import tensor_to_scale, to_fp8_saturated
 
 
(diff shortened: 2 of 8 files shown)


float8_experimental/config.py
```diff
@@ -14,3 +14,8 @@
 # this doesn't work with autocast + torch.compile + FSDP. Enabling this
 # option is useful for safety, but not strictly necessary.
 enable_pre_and_post_forward = True
+
+# If True, dynamic linear uses hooks for activation casting
```
drisspg, line 18: need to figure out if we want this

test/test_base.py
```diff
@@ -146,27 +168,59 @@ def test_linear_bias(
...
+        self._test_linear_impl(x, m_ref, linear_type, emulate, use_activation_hooks)
 
-        m = nn.Linear(32, 16, device="cuda", dtype=linear_dtype)
-        m = Float8Linear.from_float(m, emulate)
+    @pytest.mark.parametrize("emulate", [True, False])
+    @pytest.mark.parametrize("linear_type", [LinearType.DELAYED, LinearType.DYNAMIC])
+    @pytest.mark.parametrize(
+        "linear_dtype", [torch.float16, torch.bfloat16, torch.float32]
+    )
+    @pytest.mark.parametrize("use_activation_hooks", [True, False])
```
drisspg, line 183: There was some testing that was globbed together before this split the test into two

drisspg:
@bdhirsh , @wanchaol  This is a dupe and some on #170. Module hooks indeed fail with compile today for the same, multiple fake modes in env

float8_experimental/float8_dynamic_linear.py
```diff
@@ -31,13 +32,23 @@ def forward(
...
+
+
+def cast_x_to_float8_e4m3fn_pre_hook(module, args):
+    """
+    Hook to cast the incoming activation to `torch.float8_e4m3fn`
+    """
+    return module.cast_to_float8_e4m3fn(args[0])
+
+
+def cast_dldy_to_float8_e5m2_backward_pre_hook(module, grad_output):
```
wanchaol, line 46: testing my understanding: 

It looks to me that the torch.compile issue is related to the backward_prehook and tensor subclass interactions. 

But could this be solved by using `module.register_forward_hook` and then inside the forward hook we call `y = self.cast_to_float8_e5m2_bw(y)` just like the current casting?

This would solve the backward prehook issue I believe, but not sure if it would hit subclass issues (hopefully not)

wanchaol (commented):
Looks great!! I have only one comment as I thought we can avoid the backward prehook by just using a forward hook + the existing autograd function.

float8_experimental/float8_dynamic_linear.py
drisspg: ahhh great, idea unfortunately still erroring for compile..

I think that is a little cleaner, is that enough for a good interaction with DTensor?

wanchaol: Yeah I think a forward_hook might be a little cleaner and easier for DTensor to interact. We would need to actually try composing those hooks to see if there's any gap. I feel let's try land the forward_hook approach in this PR, and if we found we need `full_backward_pre_hook` instead of `forward_hook` later, we can always change it back later if needed?

wanchaol (approved):
stamp to unblock, thanks for getting this to work!

float8_experimental/float8_dynamic_linear.py
drisspg: sounds good!

Issue #201, opened 2024-02-06
Float8 + Per-Parameter FSDP


drisspg:
We will investigate fp8 all-gather for both dynamic and delayed scaling. Throughout, "fp8 all-gather" refers to moving casts to fp8 to _before_ all-gather specifically for `Float8Linear.weight`/`Float8DynamicLinear.weight`. Other parameters are all-gathered with dtype according to FSDP's mixed precision policy.

## Per-parameter FSDP with dynamic scaling
- fp8 all-gather may not be justified due to requiring all-reducing the amax (across data parallel workers) before computing scales.
    - Batching these all-reduces into one in eager requires some extra API and internals change, where the amax could not be computed inline but rather externally for many weights at once, e.g. after the optimizer step.
    - With respect to communication time, no batching could nullify the savings from fp8 all-gather in a latency-bound regime and only be helpful in a bandwidth-bound regime.
- Dynamic scaling in eager has too much CPU overhead. Compiling seems like a requirement.
    - Without fp8 all-gather, we can already compile the dynamic scaling since it is part of the module forward, similar to existing FSDP.
    - With fp8 all-gather, we have issues tracing the `Float8DynamicLinearWeightTensor` subclass added to run the fp8 all-gather: P1182470218.
- fp8 all-gather with bf16 mixed precision has different numerics than bf16 all-gather because the amax/scale is computed from the fp32 sharded parameter for the fp8 all-gather but from the bf16 unsharded parameter for the bf16 all-gather.
- On having the `Float8DynamicLinear.weight` be a tensor subclass:
    - This was to define `fsdp_pre_all_gather()` and `fsdp_post_all_gather()` methods to customize the all-gather.
    - We have to pass `foreach=True` to the optimizer to use foreach path.
    - The `Float8Tensor._data` is now of type `Float8DynamicLinearWeightTensor`. For now, this means calls (e.g. `numel()`/`view()`), go through `__torch_function__`, which adds overhead.
    - To-do: We need to make this compile-friendly and see how to save state on the tensor that can be propagated. For the latter, we currently pass `module` to `fsdp_pre_all_gather()` as a way for the method to read state.

## Per-parameter FSDP with delayed scaling
Not yet started

awgu:
## Per-parameter FSDP with dynamic scaling
- Should we all-reduce the amax before computing the scale?
- If using FSDP's bf16 mixed precision, using fp8 all-gather vs. not gives different numerics.
[...]

PR #203, opened 2024-02-07
[POC][WIP] Added fp8 all-gather

awgu:
### Overview
This PR shows the prototype for enabling fp8 all-gather for `Float8Linear.weight` and `Float8DynamicLinear.weight`. This requires changes from https://github.com/pytorch/pytorch/pull/119378.

The approach is to change the `weight` tensor into a tensor subclass that defines two methods `fsdp_pre_all_gather()` and `fsdp_post_all_gather()`. We currently prefer this approach since subclasses are the blessed approach for extending at the tensor level. However, we are evaluating the implications on both eager performance and compile.

See https://github.com/pytorch-labs/float8_experimental/issues/201 for some more notes on per-parameter FSDP and fp8.

### `torch.compile` w/o fp8 all-gather
**TL;DR** only `transformer_block.forward = torch.compile(transformer_block.forward)` works today.
|                           | Delayed Scaling                                          | Dynamic Scaling                                                      |
|---------------------------|----------------------------------------------------------|----------------------------------------------------------------------|
| Compile Transformer Block `forward` | 🙁 requires disabling amax init<br> 🙁 requires disabling pre/post-forward <br> ✅ one graph per transformer block | ✅ one graph per transformer block <br>❌ error in float8_mm if compiling output projection |
| Compile Transformer       | 🙁 requires disabling amax init<br> ❌ unexpected graph breaks                                | ❌ error in float8_mm compile                                         |

### `torch.compile` w/ fp8 all-gather
Context: Per-parameter FSDP runs all-gather in a pre-forward hook and frees parameters in a post-forward hook. Using `transformer_block.forward = torch.compile(transformer_block.forward)` does not compile the hooks, so we distinguish between two cases when doing this block-level compile: including and not including hooks.

One way to emulate including the hooks is to change per-parameter FSDP to override `forward()` instead of use hooks, in which case `transformer_block.forward` would include FSDP's pre/post-forward logic directly. We have not investigated this yet.

|                           | Delayed Scaling                                                  | Dynamic Scaling                                                                       |
|---------------------------|------------------------------------------------------------------|---------------------------------------------------------------------------------------|
| Compile Transformer Block `forward` w/o Forward Hooks | 🙁 requires disabling amax init <br>🙁 requires disabling pre/post-forward <br> ✅ one graph per transformer block         | ✅ one graph per transformer block <br>❌ error in float8_mm if compiling output projection |
| Compile Transformer       | 🙁 requires disabling amax init <br>❌ error in pre-all-gather compile | ❌ error in float8_mm compile                                                          |
| Compile Transformer Block incl. Forward Hooks |                                                                  |                                                                                       |

### DIFF
--- a/test/test_fsdp/test_fsdp_compile.py
+++ b/test/test_fsdp/test_fsdp_compile.py
@@ -1,0 +1,226 @@
+import contextlib
+from typing import Type
+
+import torch
+import torch._dynamo.testing
+import torch.distributed as dist
+from float8_experimental import config
+from float8_experimental.float8_dynamic_linear import Float8DynamicLinear
+from float8_experimental.float8_linear import Float8Linear
+from float8_experimental.float8_linear_utils import swap_linear_with_float8_linear
+from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
+from torch.distributed.fsdp.wrap import ModuleWrapPolicy
+from torch.testing._internal.common_distributed import skip_if_lt_x_gpu
+from torch.testing._internal.common_utils import run_tests
+from torch.testing._internal.distributed._tensor.common_dtensor import (
+    ModelArgs,
+    Transformer,
+    TransformerBlock,
+)
+
+
+class TestFloat8Compile(torch._dynamo.test_case.TestCase):
+    def setUp(self):
+        super().setUp()
+        from torch.testing._internal.distributed.fake_pg import FakeStore
+
+        fake_store = FakeStore()
+        dist.init_process_group(
+            "fake", store=fake_store, rank=0, world_size=self.world_size
+        )
+
+    def tearDown(self):
+        super().tearDown()
+        dist.destroy_process_group()
+
+    @property
+    def world_size(self) -> int:
+        return min(torch.cuda.device_count(), 2)
+
+    def _init_transformer_with_fp8(self, module_cls: Type, use_fp8_all_gather: bool):
+        torch.manual_seed(42)
+        args = ModelArgs(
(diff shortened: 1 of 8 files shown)


test/test_fsdp/test_fsdp_compile.py
```diff
@@ -0,0 +1,183 @@
...
+from torch.testing._internal.common_distributed import skip_if_lt_x_gpu
+from torch.testing._internal.common_utils import run_tests
+from torch.testing._internal.distributed._tensor.common_dtensor import (
+    ModelArgs,
+    Transformer,
+    TransformerBlock,
+)
+
+
+class TestFloat8Compile(torch._dynamo.test_case.TestCase):
```
awgu, line 22: The flat-parameter FSDP + `torch.compile` compatibility matrix is the same as per-parameter FSDP + `torch.compile`.

test/test_fsdp2/test_fsdp2_compile.py
```diff
@@ -0,0 +1,208 @@
...
+                    num_compiled_fns += 1
+                    fully_shard(submodule)
+            fully_shard(module)
+            module(local_inp).sum().backward()
+            self.assertEqual(cnt.frame_count, 18)  # TODO!
+            # self.assertEqual(cnt.frame_count, num_compiled_fns)
+
+        # Compile each transformer block separately with amax init disabled and
+        # pre/post-forward disabled
+        with self.enable_amax_init(False), self.enable_pre_and_post_forward(False):
```
awgu, line 165: With _both_ amax init disabled and pre/post-forward disabled, we can get the expected number of compiled functions.

awgu:
Closing to re-open as ghstack

Issue #205, opened 2024-02-12
swap_linear_with_float8_linear does not pipe skip_fqn_list, cur_fqn to recursion


yitzhaklevi:
Hey, 
as you can [see here](https://github.com/pytorch-labs/float8_experimental/blob/main/float8_experimental/float8_linear_utils.py#L94) those arguments are not piped during the recursive call.
effectively only top level model.named_children() can be ignored.

awgu:
Some other enhancement to `swap_linear_with_float8_linear` (somewhat inspired by learnings from FSDP):
- The swap function should return `nn.Module` so that it can be called on a single `nn.Linear`.
- nit: `model` should be renamed to `module` because the swap function can be called on any module (may not be the overall model).

Issue #206, opened 2024-02-12
swap_linear_with_float8_linear does not swap torch.nn.Linear if it's the model itself


yitzhaklevi:
`
# this would do nothing
model1 = torch.nn.Linear([16, 16])
swap_linear_with_float8_linear(model1, Float8Linear)

# this would work
model2 = torch.nn.Sequential(torch.nn.Linear([16, 16]))
swap_linear_with_float8_linear(model2, Float8Linear)
`
the reason is due to the fact that model1 has no child modules, 
(the use case is quite odd, but it would be useful to change the function signature since it's not working as expected)

awgu:
https://github.com/pytorch-labs/float8_experimental/pull/208

PR #208, opened 2024-02-12
Improved `swap_linear_with_float8_linear`

awgu:
This PR improves `swap_linear_with_float8_linear` to fix https://github.com/pytorch-labs/float8_experimental/issues/205 and https://github.com/pytorch-labs/float8_experimental/issues/206.

**Test Plan**
Added unit tests for `swap_linear_with_float8_linear` to `test_base.py`.

```
./test/test_fsdp.sh
pytest test/test_base.py
```

### DIFF
--- a/float8_experimental/float8_linear_utils.py
+++ b/float8_experimental/float8_linear_utils.py
@@ -5,9 +5,11 @@
 # LICENSE file in the root directory of this source tree.
 import copy
 from enum import auto, Enum
+from typing import List, Optional, Type
 
 import torch
 import torch.distributed as dist
+import torch.nn as nn
 from float8_experimental.float8_dynamic_linear import Float8DynamicLinear
 from float8_experimental.float8_linear import Float8Linear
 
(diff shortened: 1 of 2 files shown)


float8_experimental/float8_linear_utils.py
```diff
@@ -66,32 +68,53 @@ def _update_history_with_new_amax(new_amax, amax_history):
...
-    module,
-    emulate=False,
-    skip_fqn_list=None,
-    cur_fqn="",
-):
+    module: nn.Module,
+    module_cls: Type[nn.Module],
+    emulate: bool = False,
+    skip_fqn_list: Optional[List[str]] = None,
+    cur_fqn: str = "",
```
awgu, line 75: @drisspg Do you have historical context on `cur_fqn`?

In particular, did users want to be able to pass `skip_fqn_list` as module FQNs that are "globally" prefixed from the root module, while only call `swap_linear_with_float8_linear` on a submodule?
```
Example:
skip_fqn_list = ["dense_arch.mlp.lin1", "dense_arch.mlp.lin2"]
swap_linear_with_float8_linear(model.dense_arch, ..., skip_fqn_list, cur_fqn="dense_arch")
```

Or, was `cur_fqn` always expected to be `""` upon the user's call to `swap_linear_with_float8_linear`?

If it is the latter case, then personally I would prefer removing this argument as to not leak arguments used for recursion to the user.

awgu (commented):
@drisspg Do you know how to run the existing tests?

Following the README with `./test/test_fsdp.py`, I get:
```
ModuleNotFoundError: No module named 'float8_experimental'
```

I try to use `python -m pytest` to work around these, but some of these tests like `test_fsdp_compile.py` use fire.

drisspg:
For the first question: @awgu  did you install the float8 exp -> `pip install -e .`

awgu:
@drisspg Amazing! Thank you! `pip install -e .` fixed it. (I should have read the README better 😢 )

float8_experimental/float8_linear_utils.py
```diff
@@ -66,32 +68,55 @@ def _update_history_with_new_amax(new_amax, amax_history):
...
-            child, torch.nn.Linear
-        ):
-            new_child = module.from_float(child, emulate)
-            setattr(model, name, new_child)
-        else:
-            swap_linear_with_float8_linear(child, module, emulate)
+    module_names_to_skip = set(skip_fqn_list or [])
+    if isinstance(module, nn.Linear):
+        if cur_fqn not in module_names_to_skip:
+            assert len(list(module.children())) == 0, f"{module}"
```
drisspg, line 93: Maybe a better error message

drisspg:
One follow up I had been thinking about was to skip linear modules that don't have the correct alignment constraints, k%16 ==0

PR #209, opened 2024-02-12
Add tests for scaled_mm + DTensor

wanchaol:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #210
* __->__ #209

This PR adds test to test the scaled_mm op working with DTensor.

Currently using the same style of test runner
compared to other tests in float8_experimental. We shall consider
switching to DTensorTestBase or MultiProcessTestCase in the repo but
didn't do it in this PR

Differential Revision: [D53731879](https://our.internmc.facebook.com/intern/diff/D53731879)

PR #210, opened 2024-02-12
support DTensor redistribute to replicate (allgather)

wanchaol:
* __->__ #210
* #209

This PR adds support for fp8tensor allgather and make DTensor(Shard(0))
that holds a Float8Tensor to be able to do redistribute to replication

Differential Revision: [D53731878](https://our.internmc.facebook.com/intern/diff/D53731878)

### DIFF
--- a/float8_experimental/float8_ops.py
+++ b/float8_experimental/float8_ops.py
@@ -13,6 +13,7 @@ from float8_experimental.float8_utils import is_row_major
 from torch.utils._pytree import tree_map
 
 aten = torch.ops.aten
+c10d_functional = torch.ops.c10d_functional
 FLOAT8_OPS_TABLE: Dict[Any, Any] = {}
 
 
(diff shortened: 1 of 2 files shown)


float8_experimental/float8_ops.py
```diff
@@ -140,3 +141,29 @@ def autocast_to_copy(aten_op, args, kwargs=None):
...
         args[0]._data, args[0]._scale, kwargs["dtype"], args[0]._emulate
     )
+
+
+@implements([c10d_functional.all_gather_into_tensor.default])
+def allgather_fp8(aten_op, args, kwargs=None):
+    """
+    override funcol with FP8 handling
+    """
+    fp8_input = args[0]
```
awgu, line 151: nit: I am not familiar, but what if the user passes only via `kwargs`?
Similarly, do we need `if kwargs is None: kwargs = {}`?

wanchaol: I think our python arg parser would parse the python inputs using the operator schema, so even if user pass as kwarg, it would be passed as arg if the operator does not have this kwarg.

For this kwarg handling
```
if kwargs is None: kwargs = {}
```
iirc passing kwarg directly down to the aten call should not have a problem, I think some of the someclasses (i.e. DTensor) handle kwarg this way mainly because of avoiding type annotation and isinstance checks of None values.

awgu:
For linting, I have ufmt installed in my VSCode, so I have just been doing `Format Document`.

float8_experimental/float8_ops.py
```diff
@@ -140,3 +141,29 @@ def autocast_to_copy(aten_op, args, kwargs=None):
...
+    assert isinstance(fp8_input, Float8Tensor)
```
drisspg, line 152: maybe a helpful message, you would be surprised what happends during torch.compile lol

float8_experimental/float8_ops.py
```diff
@@ -140,3 +141,29 @@ def autocast_to_copy(aten_op, args, kwargs=None):
...
+
+    fp8_data = fp8_input._data
+    fp8_data = fp8_data.view(torch.int8)
```
drisspg, line 155: curious q: is there any difference between this and torch.uint8?

float8_experimental/float8_ops.py
```diff
@@ -140,3 +141,29 @@ def autocast_to_copy(aten_op, args, kwargs=None):
...
+    fp8_data = fp8_data.contiguous()
```
drisspg, line 156: should this be an assert is.contiguous .. I guess not I think the _make_wrapper_subclass consumes the input data strides so should be fine

float8_experimental/float8_ops.py
awgu: JFYI @wanchaol : I was using `uint8` for fp8 all-gather work because `uint8` felt more traditional for "the raw bytes", but I think it should not matter much.

drisspg (approved):
Looks good!

float8_experimental/float8_ops.py
wanchaol: I was following https://github.com/pytorch-labs/float8_experimental/blob/main/float8_experimental/distributed_utils.py#L49 but torch.uint8 make more sense, switching over

wanchaol:
@wanchaol has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D53731878).

wanchaol:
@wanchaol has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D53731878).

PR #211, opened 2024-02-12
Updates to sync_float_amax_history

drisspg:
# Summary

Update docs, make sure this is friendly to dynamo

### Perf

PyTorch Version | float8 Version | Eager Iterations per Second | Compile
-- | -- | -- | --
Nightly | Main | 1.15 it/s |  2.10 it/s
Nightly | This PR | 1.16 it/s | 2.27 it/s

Trace | Compile URL | Eager
-- | -- | --
This PR | https://fburl.com/753ztao4 |  https://fburl.com/34yftzao
Main |  https://fburl.com/a0gh9iof |  https://fburl.com/u9c4ilmp

#### Commit 1
- [x] We previously had an `fp8_classes` argument that would be passed in, this was to enable working with the separate TP/SP classes, since we plan to have Dtensor be the solution I am removing for now.
- [x] I put the child.amax_and_scale_synced module mutation under the enable_amax_init flag, this seemed to be causing graphbreaks cause of the module mutation

#### Commit 2

- [x] We previously had all the history buffers be scaler tensors. This meant that to construct the combined tensor we needed to call torch.Tensor which was causing a HtoD sync under torch.compile. I needed to added a single dimension of size 1 and pipe that through all the places.
- [x] Note that this meant we needed to update the to_hp to send back to original precision because [line](https://github.com/pytorch-labs/float8_experimental/pull/211/commits/f3630d05881f324eecb8e6e1837a3f487e77a7ff#diff-94b99416a4df6d75c548de330c1f71505e830b3afff114213d131cf2620597efR57-R59) the scale upcasts the _data tensor

#### Commit 3
- [x] Rewrote the sync function to do the torch.roll() on all the histories at once - side note not sure if this is more expensive than to clones since we really dont care about the wrapping behavior
- [x] Same for generating the new scales from the grouped histories

##### Things to do
- There is still two loops and those are for mutating the the actual module values, not sure if there is another way around this..
- Going to try the functional collectives

### DIFF
--- a/float8_experimental/float8_linear_utils.py
+++ b/float8_experimental/float8_linear_utils.py
@@ -196,25 +196,25 @@ def sync_float8_amax_and_scale_history(model: torch.nn.Module, fp8_layers=None) 
         return
 
     # Loop over all fp8 layers and grab the needed tensors
-    fp8_amax_x_tensor_list = []
-    fp8_amax_w_tensor_list = []
-    fp8_amax_dL_dY_tensor_list = []
+    fp8_amax_x_tensor_list = [None] * len(fp8_layers)
+    fp8_amax_w_tensor_list = [None] * len(fp8_layers)
+    fp8_amax_dL_dY_tensor_list = [None] * len(fp8_layers)
 
-    fp8_x_amax_history_stack = []
-    fp8_w_amax_history_stack = []
-    fp8_dL_dY_amax_history_stack = []
+    fp8_x_amax_history_stack = [None] * len(fp8_layers)
+    fp8_w_amax_history_stack = [None] * len(fp8_layers)
+    fp8_dL_dY_amax_history_stack = [None] * len(fp8_layers)
 
     x_dtypes = set()
     scale_fn_recipes = set()
 
-    for child in fp8_layers:
-        fp8_amax_x_tensor_list.append(child.fp8_amax_x)
-        fp8_amax_w_tensor_list.append(child.fp8_amax_w)
-        fp8_amax_dL_dY_tensor_list.append(child.fp8_amax_dL_dY)
+    for idx, child in enumerate(fp8_layers):
+        fp8_amax_x_tensor_list[idx] = child.fp8_amax_x
+        fp8_amax_w_tensor_list[idx] = child.fp8_amax_w
+        fp8_amax_dL_dY_tensor_list[idx] = child.fp8_amax_dL_dY
 
-        fp8_x_amax_history_stack.append(child.fp8_amax_history_x)
-        fp8_w_amax_history_stack.append(child.fp8_amax_history_w)
-        fp8_dL_dY_amax_history_stack.append(child.fp8_amax_history_dL_dY)
+        fp8_x_amax_history_stack[idx] = child.fp8_amax_history_x
+        fp8_w_amax_history_stack[idx] = child.fp8_amax_history_w
+        fp8_dL_dY_amax_history_stack[idx] = child.fp8_amax_history_dL_dY
 
         x_dtypes.add(child.last_seen_input_dtype)
         scale_fn_recipes.add(child.recipe.scale_fn_name)
--- a/float8_experimental/float8_linear_utils.py
+++ b/float8_experimental/float8_linear_utils.py
@@ -155,9 +155,7 @@ def get_float8_layers(model: torch.nn.Module):
 
     # Get all fp8 layers and tensors
     fp8_layers = [
-        child
-        for name, child in model.named_modules()
-        if isinstance(child, Float8Linear)
+        child for _, child in model.named_modules() if isinstance(child, Float8Linear)
     ]
 
     return fp8_layers
@@ -195,9 +193,36 @@ def sync_float8_amax_and_scale_history(model: torch.nn.Module, fp8_layers=None) 
         )
         return
 
-    fp8_amax_x_tensor_list = [child.fp8_amax_x for child in fp8_layers]
-    fp8_amax_w_tensor_list = [child.fp8_amax_w for child in fp8_layers]
-    fp8_amax_dL_dY_tensor_list = [child.fp8_amax_dL_dY for child in fp8_layers]
+    # Loop over all fp8 layers and grab the needed tensors
+    fp8_amax_x_tensor_list = []
+    fp8_amax_w_tensor_list = []
+    fp8_amax_dL_dY_tensor_list = []
+
+    fp8_x_amax_history_stack = []
+    fp8_w_amax_history_stack = []
+    fp8_dL_dY_amax_history_stack = []
+
+    x_dtypes = set()
+    scale_fn_recipes = set()
+
+    for child in fp8_layers:
+        fp8_amax_x_tensor_list.append(child.fp8_amax_x)
+        fp8_amax_w_tensor_list.append(child.fp8_amax_w)
+        fp8_amax_dL_dY_tensor_list.append(child.fp8_amax_dL_dY)
+
+        fp8_x_amax_history_stack.append(child.fp8_amax_history_x)
+        fp8_w_amax_history_stack.append(child.fp8_amax_history_w)
+        fp8_dL_dY_amax_history_stack.append(child.fp8_amax_history_dL_dY)
+
+        x_dtypes.add(child.last_seen_input_dtype)
+        scale_fn_recipes.add(child.recipe.scale_fn_name)
+
+    # TODO This way to get the activation dtype is not ideal
+    assert len(x_dtypes) == 1, "All layers must have the same last seen input_dtype"
+    x_dtype = next(iter(x_dtypes))
+
+    assert len(scale_fn_recipes) == 1, "All layers must have the same scale_fn recipe"
+    scale_fn_recipe = next(iter(scale_fn_recipes))
 
     assert (
         len(fp8_amax_x_tensor_list)
@@ -216,40 +241,23 @@ def sync_float8_amax_and_scale_history(model: torch.nn.Module, fp8_layers=None) 
         dist.all_reduce(fp8_amax_dL_dY_tensor, op=dist.ReduceOp.MAX)
 
         # Reassign the reduced amax values to the original tensors
-        for idx in range(len(fp8_layers)):
-            child = fp8_layers[idx]
-            # Do we need this extra clone?
-            child.fp8_amax_x.copy_(fp8_amax_x_tensor[idx].clone())
-            child.fp8_amax_w.copy_(fp8_amax_w_tensor[idx].clone())
-            child.fp8_amax_dL_dY.copy_(fp8_amax_dL_dY_tensor[idx].clone())
+        for idx, child in enumerate(fp8_layers):
+            child.fp8_amax_x.copy_(fp8_amax_x_tensor[idx])
+            child.fp8_amax_w.copy_(fp8_amax_w_tensor[idx])
+            child.fp8_amax_dL_dY.copy_(fp8_amax_dL_dY_tensor[idx])
 
     # We create two stacked tensors, one for the amax history and one for the current scales
     fp8_amax_x_tensors = torch.vstack(fp8_amax_x_tensor_list)
     fp8_amax_w_tensors = torch.vstack(fp8_amax_w_tensor_list)
     fp8_amax_dL_dY_tensors = torch.vstack(fp8_amax_dL_dY_tensor_list)
 
-    fp8_x_amax_history_stack = torch.stack(
-        [child.fp8_amax_history_x for child in fp8_layers]
-    )
-    fp8_w_amax_history_stack = torch.stack(
-        [child.fp8_amax_history_w for child in fp8_layers]
-    )
-    fp8_dL_dY_amax_history_stack = torch.stack(
-        [child.fp8_amax_history_dL_dY for child in fp8_layers]
-    )
+    fp8_x_amax_history_stack = torch.vstack(fp8_x_amax_history_stack)
+    fp8_w_amax_history_stack = torch.vstack(fp8_w_amax_history_stack)
+    fp8_dL_dY_amax_history_stack = torch.vstack(fp8_dL_dY_amax_history_stack)
 
     _update_history_stack(fp8_amax_x_tensors, fp8_x_amax_history_stack)
     _update_history_stack(fp8_amax_w_tensors, fp8_w_amax_history_stack)
     _update_history_stack(fp8_amax_dL_dY_tensors, fp8_dL_dY_amax_history_stack)
-
-    # TODO This way to get the activation dtype is not ideal
-    x_dtypes = {child.last_seen_input_dtype for child in fp8_layers}
-    assert len(x_dtypes) == 1, "All layers must have the same last seen input_dtype"
-    x_dtype = next(iter(x_dtypes))
-
-    scale_fn_recipes = {child.recipe.scale_fn_name for child in fp8_layers}
-    assert len(scale_fn_recipes) == 1, "All layers must have the same scale_fn recipe"
-    scale_fn_recipe = next(iter(scale_fn_recipes))
 
     # We are not reading the
     new_x_scales = amax_history_to_scale_stack(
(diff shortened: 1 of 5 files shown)


float8_experimental/float8_linear_utils.py
```diff
@@ -135,84 +159,89 @@ def sync_float8_amax_and_scale_history(model: torch.nn.Module, fp8_layers=None)
...
+            child.fp8_amax_x.copy_(fp8_amax_x_tensor[idx].clone())
+            child.fp8_amax_w.copy_(fp8_amax_w_tensor[idx].clone())
+            child.fp8_amax_dL_dY.copy_(fp8_amax_dL_dY_tensor[idx].clone())
+
+    # We create two stacked tensors, one for the amax history and one for the current scales
+    fp8_amax_x_tensors = torch.vstack(fp8_amax_x_tensor_list)
+    fp8_amax_w_tensors = torch.vstack(fp8_amax_w_tensor_list)
+    fp8_amax_dL_dY_tensors = torch.vstack(fp8_amax_dL_dY_tensor_list)
+
+    fp8_x_amax_history_stack = torch.stack(
```
drisspg, line 204: should probs use vstack here again just to be sure

awgu (commented):
I left some initial comments just scanning through the code.

float8_experimental/float8_linear_utils.py
```diff
@@ -120,95 +148,101 @@ def sync_float8_amax_and_scale_history(
...
-            device="cuda",
-            requires_grad=False,
-        )
-        fp8_amax_dL_dY_tensor = torch.tensor(
-            [child.fp8_amax_dL_dY for child in fp8_layers],
-            dtype=torch.float32,
-            device="cuda",
-            requires_grad=False,
-        )
+        # Combine all the amax tensors into one tensor and reduce it
```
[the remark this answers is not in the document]
drisspg: ohhh great call!

float8_experimental/float8_linear_utils.py
```diff
@@ -120,95 +148,101 @@ def sync_float8_amax_and_scale_history(
...
+        fp8_amax_dL_dY_tensor = torch.cat(fp8_amax_dL_dY_tensor_list)
+
         dist.all_reduce(fp8_amax_x_tensor, op=dist.ReduceOp.MAX)
         dist.all_reduce(fp8_amax_w_tensor, op=dist.ReduceOp.MAX)
         dist.all_reduce(fp8_amax_dL_dY_tensor, op=dist.ReduceOp.MAX)
 
-    for idx in range(len(fp8_layers)):
-        child = fp8_layers[idx]
+        # Reassign the reduced amax values to the original tensors
+        for idx in range(len(fp8_layers)):
```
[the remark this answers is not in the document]
drisspg: This is the exact func I was looking for, does it not work with compile?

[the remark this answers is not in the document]
awgu: I have not experimented with it much, so I think it is worth trying out.

The one time I did try it, maybe since I only compiled a small function, the compiled function was worse than eager due to some expensive setup to begin.

float8_experimental/float8_linear_utils.py
```diff
@@ -120,95 +146,110 @@ def sync_float8_amax_and_scale_history(
...
-        child.fp8_scale_w.copy_(new_scale)
-        new_scale = amax_history_to_scale(
-            child.fp8_amax_history_dL_dY,
-            torch.float8_e5m2,
-            x_dtype,
-            child.recipe.scale_fn_name,
-        )
-        child.fp8_scale_dL_dY.copy_(new_scale)
+    # We create two stacked tensors, one for the amax history and one for the current scales
+    fp8_amax_x_tensors = torch.vstack(fp8_amax_x_tensor_list)
```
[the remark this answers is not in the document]
drisspg: My thought process here is that we create a python list that points to tensors. We than create a new "cat" tensor that is constructed from the list at state 1. We all reduce and then inplace modify the tensors in that list. If we don't re-construct another tensor, then we will have stale data.

I used cat cause I assumed that all-reduce wants flat things and then vstack cause I wanted, `N x History Length`

does that make sense or am I being overly cautious

float8_experimental/float8_linear_utils.py
```diff
@@ -120,95 +146,110 @@ def sync_float8_amax_and_scale_history(
...
-
-        #
-        # 2. adds the `amax` values to history
-        #
-        _update_history_with_new_amax(child.fp8_amax_x, child.fp8_amax_history_x)
-        _update_history_with_new_amax(child.fp8_amax_w, child.fp8_amax_history_w)
-        _update_history_with_new_amax(
-            child.fp8_amax_dL_dY, child.fp8_amax_history_dL_dY
-        )
+        # Reassign the reduced amax values to the original tensors
```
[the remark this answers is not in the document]
drisspg: Yeah and FWIW TE does something similar. I think it does make it harder to reason about what is going

float8_experimental/float8_linear_utils.py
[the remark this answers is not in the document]
drisspg: Tried, out it currently does not work with AsyncTensorCollective subclass

float8_experimental/float8_tensor.py
```diff
@@ -54,7 +54,9 @@ class FromFloat8ConstrFunc(torch.autograd.Function):
 
     @staticmethod
     def forward(ctx, tensor):
-        return tensor._data.to(tensor._orig_dtype) / tensor._scale
+        return (tensor._data.to(tensor._orig_dtype) / tensor._scale).to(
```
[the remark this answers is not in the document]
drisspg: I ended up fixing this probs a bug I fixed along the way

float8_experimental/float8_linear_utils.py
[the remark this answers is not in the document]
bdhirsh: fwiw - there's some precedent for this kind of thinking working in compile, I think (r.e. shampoo)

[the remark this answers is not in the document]
drisspg: I think if we want do this that should be a follow up TBH and should be very data driven, what do you guys think?

But the path down would be, coalescing all amax metadata into single tensors on an individual Float8Linear and then if need be coalescing all of these into GLOBAL tensors, which personally I dont like but maybe warranted for perf.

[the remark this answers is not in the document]
drisspg: what would be really great is to get cuda-graphs working but alas:

```Shell
skipping cudagraphs due to mutaton on input. Found from : 
   File "/home/drisspg/miniconda3/envs/nightly/lib/python3.10/site-packages/torch/_dynamo/external_utils.py", line 25, in inner
    return fn(*args, **kwargs)
  File "/home/drisspg/miniconda3/envs/nightly/lib/python3.10/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/home/drisspg/meta/float8_experimental/float8_experimental/float8_linear_utils.py", line 226, in sync_float8_amax_and_scale_history
    child.fp8_amax_x.copy_(reduced_fp8_amax_tensor[idx])
    
   
    ```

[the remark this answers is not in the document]
Comment (drisspg) [reply float8_experimental/float8_linear_utils.py]: Spoke with @eellison and he says that mutation on module buffers should be supported, and will try and look at this sometime in the future

File: float8_experimental/float8_linear_utils.py [hunk 7]
```diff
@@ -65,6 +72,26 @@ def _update_history_with_new_amax(new_amax, amax_history):
...
+) -> torch.Tensor:
+    """
+    Updates `amax_history` (the last N cur_amax values) inplace with the value
+    of `new_amax`.
+
+    Args:
+        new_amax (torch.Tensor): The new amax value to add to the history. (n_amaxes, 1)
+        amax_history_stack (torch.Tensor): The history of amax values. (n_amaxes, history_length)
+    """
+    assert amax_history_stack.dim() == 2, "amax_history_stack must be 2D"
```
awgu, line 86: nit: Include the `amax_history_stack.shape` in the assert error message? (similarly for below assert)

float8_experimental/float8_linear_utils.py
```diff
@@ -94,21 +121,22 @@ def swap_linear_with_float8_linear(
...
+def get_float8_layers(model: torch.nn.Module):
+    """Iterates through the model and returns all the Float8Linear layers.
+    Args:
+        model (torch.nn.Module): The model to look for Float8Linear layers in.
+    """
 
     # Get all fp8 layers and tensors
     fp8_layers = [
-        child for name, child in model.named_modules() if isinstance(child, fp8_classes)
+        child for _, child in model.named_modules() if isinstance(child, Float8Linear)
```
awgu, line 132: nit: 
```suggestion
        child for child in model.modules() if isinstance(child, Float8Linear)
```

float8_experimental/float8_linear_utils.py
```diff
@@ -120,95 +148,117 @@ def sync_float8_amax_and_scale_history(
 
     TODO(future): design the UX for this (context manager, etc)
 
+    PERFORMANCE NOTE:
+        When you can it is much more efficient to call te get_float8_layers once a
```
awgu, line 152: nit:
```suggestion
        When you can, it is much more efficient to call get_float8_layers once at
```

float8_experimental/float8_linear_utils.py
```diff
@@ -120,95 +148,117 @@ def sync_float8_amax_and_scale_history(
...
+
+        fp8_x_amax_history_stack[idx] = child.fp8_amax_history_x
+        fp8_w_amax_history_stack[idx] = child.fp8_amax_history_w
+        fp8_dL_dY_amax_history_stack[idx] = child.fp8_amax_history_dL_dY
+
+        x_dtypes.add(child.last_seen_input_dtype)
+        scale_fn_recipes.add(child.recipe.scale_fn_name)
+
+    # TODO This way to get the activation dtype is not ideal
+    assert len(x_dtypes) == 1, "All layers must have the same last seen input_dtype"
```
awgu, line 196: nit: Include `x_dtypes` in the assert message? (but then may need to be careful of eagerly evaluating the `__repr__`, so may need to change to an `if len(x_dtypes) != 1: raise ...`)

similarly for below -- including the info that caused the failing assert might be helpful

float8_experimental/float8_linear_utils.py
```diff
@@ -120,95 +148,117 @@ def sync_float8_amax_and_scale_history(
...
-        #
-        _update_history_with_new_amax(child.fp8_amax_x, child.fp8_amax_history_x)
-        _update_history_with_new_amax(child.fp8_amax_w, child.fp8_amax_history_w)
-        _update_history_with_new_amax(
-            child.fp8_amax_dL_dY, child.fp8_amax_history_dL_dY
+        all_reduced_amax_tensor = all_reduce(
+            all_amax_tensors, "MAX", list(range(dist.get_world_size()))
         )
+        if isinstance(all_reduced_amax_tensor, AsyncCollectiveTensor):
+            all_reduced_amax_tensor = all_reduced_amax_tensor.wait()
```
awgu, line 217: Aside: It seems like a bug on `AsyncCollectiveTensor` if it does not make sure to call `wait()` when you `torch.split` it. 🤔

float8_experimental/float8_linear_utils.py
```diff
@@ -120,95 +148,117 @@ def sync_float8_amax_and_scale_history(
...
-            child.recipe.scale_fn_name,
-        )
-        child.fp8_scale_dL_dY.copy_(new_scale)
+        (
+            reduced_fp8_amax_tensor,
+            reduced_fp8_amax_w_tensor,
+            reduced_fp8_amax_dL_dY_tensor,
+        ) = torch.split(all_reduced_amax_tensor, len(fp8_amax_x_tensor_list))
+
+        # TODO foreach is not supported with AsyncCollectiveTensor
```
awgu, line 225: If we `wait` on it, would it still be an `AsyncCollectiveTensor`?

float8_experimental/float8_linear_utils.py
```diff
@@ -120,95 +148,117 @@ def sync_float8_amax_and_scale_history(
...
+
+    fp8_x_amax_history_stack = torch.vstack(fp8_x_amax_history_stack)
+    fp8_w_amax_history_stack = torch.vstack(fp8_w_amax_history_stack)
+    fp8_dL_dY_amax_history_stack = torch.vstack(fp8_dL_dY_amax_history_stack)
+
+    _update_history_stack(fp8_amax_x_tensors, fp8_x_amax_history_stack)
+    _update_history_stack(fp8_amax_w_tensors, fp8_w_amax_history_stack)
+    _update_history_stack(fp8_amax_dL_dY_tensors, fp8_dL_dY_amax_history_stack)
+
+    # We are not reading the
```
awgu, line 244: nit: unfinished comment

float8_experimental/float8_linear_utils.py
drisspg: Yeah I need to update this message, the second arg isn't a list any more but rather a  tensor so we would need to unbind before for_eaching. I profiled and the unbind isn't worth the foreach. Will update the comment though

float8_experimental/float8_linear_utils.py
bdhirsh: Hmm...

I actually tried to add an optimization to AsyncCollectiveTensor a while ago: https://github.com/pytorch/pytorch/pull/105240. Where the thought behind it was:

(1) if we have a collective in-flight, we only need to sync the collective if the **data** needs to be used in an operation

(2) view ops generally don't need to access the data of their inputs, only the metadata.

So that PR "delays" syncs when there are view ops called on AsyncCollectiveTensor (e.g. `aten.split()`).

drisspg: Indeed, I was working through this with @wanchaol and without the all_reduced_amax_tensor was garbage...

bdhirsh: @drisspg I'm curious why the manual sync here is needed - does it paper over some underlying bug?

awgu: Interestingly, I was not able to repro the issue: P1184812077

`_foreach_copy_` and functional all-reduce without `.wait()` seemed to be fine, but I might not be repro-ing correctly.

drisspg: hmmm let me try again then

drisspg: You are right... however:
```Shell
    return self.__get_result()
  File "/home/drisspg/miniconda3/envs/dev/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
torch._dynamo.exc.BackendCompilerFailed: backend='inductor' raised:
CompilationError: at 10:30:def triton_(in_ptr0, in_ptr1, in_ptr2, in_ptr3, out_ptr0, out_ptr1, out_ptr3, out_ptr4, out_ptr5, out_ptr7, out_ptr8, out_ptr9, out_ptr11):
    xpid = tl.program_id(0)
    XBLOCK: tl.constexpr = 1024
    if xpid >= 0 and xpid < 1:
        xpid_offset = xpid - 0
        xnumel = 1
        xoffset = xpid_offset * XBLOCK
        xindex = xoffset + tl.arange(0, XBLOCK)[:, None]
        xmask = xindex < xnumel
        rindex = tl.arange(0, RBLOCK)[None, :]
                              ^
NameError('RBLOCK is not defined')

Set TORCH_LOGS="+dynamo" and TORCHDYNAMO_VERBOSE=1 for more information


You can suppress this exception and fall back to eager by setting:
    import torch._dynamo
    torch._dynamo.config.suppress_errors = True
```
 compiling the sync function now causes an inductor fault... so I will create another PR and investigate there but otherwise I think thats everything for this one

drisspg: https://github.com/pytorch-labs/float8_experimental/pull/213

float8_experimental/float8_linear_utils.py
```diff
@@ -147,95 +165,123 @@ def sync_float8_amax_and_scale_history(
 
     TODO(future): design the UX for this (context manager, etc)
 
+    PERFORMANCE NOTE:
+        When you can, it is much more efficient to call get_float8_layers once at
+        the beginning of the training loop and pass the result to this function.
+        Because of how this interacts with torch.compile
```
awgu, line 171: For my understanding, how does passing `float8_layers` affect `torch.compile`?

awgu (approved):
LGTM!

float8_experimental/float8_linear_utils.py
drisspg: honestly I wrote this at the beginning of the PR and now I am not sure that this is true, but I found the get_attr was causing graph breaks but that was a few iterations ago and reading the graph_breaks logs can be a lil hard to decipher some times

PR #214, opened 2024-02-14
Added `use_activation_hooks: bool` to swap

awgu:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #214

This PR exposes `use_activation_hooks` to `swap_linear_with_float8_linear`. I am not sure if we want this long-term. Are we going to consolidate to _always_ using activation hooks in the future (because we need it for DTensor (?))?

**Test Plan**
```
./test/test_everything.sh
```

Differential Revision: [D53780333](https://our.internmc.facebook.com/intern/diff/D53780333)

PR #217, opened 2024-02-15
Removed `module` arg from `fsdp_pre_all_gather`

awgu:
* __->__ #217
* #216
* #215
* #214
* #219
* #220

On the discussion of whether fp8 all-gather is viable _without_ compiling the pre-all-gather cast to fp8, this PR would add _more_ CPU overhead due to the `__torch_function__` override, making it less viable.

### DIFF
--- a/float8_experimental/float8_dynamic_linear.py
+++ b/float8_experimental/float8_dynamic_linear.py
@@ -184,8 +184,8 @@ class Float8DynamicLinearWeightTensor(torch.Tensor):
         *,
         out: Optional[torch.Tensor] = None,
     ) -> Union[Tuple[Float8Tensor, Tuple[torch.Tensor, ...]], None]:
-        data, = all_gather_outputs
-        scale, = metadata
+        (data,) = all_gather_outputs
+        (scale,) = metadata
         if out is not None:
             out = cast(Float8Tensor, out)
             assert (
--- a/float8_experimental/float8_dynamic_linear.py
+++ b/float8_experimental/float8_dynamic_linear.py
@@ -7,13 +7,16 @@
 A wrapper around a `torch.nn.Linear` module which does fp8 compute.
 """
 
-from typing import Any, cast, Optional, Tuple, Union
+import functools
+from typing import Any, Callable, cast, Optional, Tuple, Union
 
 import torch
 import torch.nn as nn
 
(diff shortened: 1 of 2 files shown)


float8_experimental/float8_linear.py
```diff
@@ -232,7 +232,7 @@ def cast_w_to_float8(
             self.fp8_scale_w,
             scale_fn_name,
             torch.float8_e4m3fn,
-            is_amax_initialized,
+            self.is_amax_initialized,
```
awgu, line 235: We have to push this `is_amax_initialized` into the cast function since the `Float8LinearWeightTensor` subclass cannot keep a reference to the bool.

float8_experimental/float8_linear.py
```diff
@@ -344,12 +346,32 @@ def from_float(
...
+    def __torch_function__(cls, func, types, args=(), kwargs=None):
+        kwargs = kwargs or {}
+
+        def wrap(cast_fn: Callable, emulate: bool, o: Any):
+            if isinstance(o, torch.Tensor) and not isinstance(o, cls):
+                return cls(o, cast_fn, emulate)
+            return o
+
+        with torch._C.DisableTorchFunctionSubclass():
+            if isinstance(args[0], cls):
```
awgu, line 367: The heuristic here is just to propagate the `Float8LinearWeightTensor` as long as it is the first argument.

float8_experimental/float8_dynamic_linear.py
```diff
@@ -142,12 +147,34 @@ def from_float(
 
 
 class Float8DynamicLinearWeightTensor(torch.Tensor):
-    # TODO: Remove `module` arg, save state on subclass, and propagate it.
-    def fsdp_pre_all_gather(
-        self, module: nn.Module
```
awgu, line 147: The problem statement is that in order to implement the pre-all-gather transform, the subclass needs some state (e.g. mainly emulate but preferably the cast function as well). In the first prototype, I shortcutted by passing `module` into `fsdp_pre_all_gather()` so that the transform could read state off the module instead of storing it on the subclass itself.

However, the morally right thing (from a design perspective) should be to put that state on the subclass. I wonder though, how does this kind of  `__torch_function__` subclass interact with `torch.compile` today. cc: @bdhirsh

PR #219, opened 2024-02-16
Used functional all-reduce for amax reduction

awgu:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #217
* #216
* #215
* #214
* __->__ #219
* #220

Dynamo cannot remap the in-place max all-reduce to its functional equivalent on PyTorch `main` branch.
https://github.com/pytorch/pytorch/issues/120082

PR #220, opened 2024-02-16
Set `amax_and_scale_synced` unconditionally

awgu:
* #217
* #216
* #215
* #214
* #219
* __->__ #220

`Float8Linear.amax_and_scale_synced` is orthogonal to `config.enable_amax_init`.
- `enable_amax_init = False` skips the lazy initialization of amaxes and scales. Without this, the 1st iteration's amaxes will be the default (`E4M3_MAX_POS`).
- `amax_and_scale_synced` is checked every pre-forward regardless of `config.enable_amax_init`.

https://github.com/pytorch-labs/float8_experimental/blob/956195bbc3983594308e7ccd64bab204c457a7a8/float8_experimental/float8_linear.py#L266-L273


drisspg:
So I added this because this module mutation was causing a graph break, hmmm

PR #224, opened 2024-02-22
Insure DTensor wraps inner float8 Tensors

drisspg:


float8_experimental/float8_ops.py
```diff
@@ -78,9 +79,16 @@ def preprocess_addmm(a: Float8Tensor, b: Float8Tensor):
 
 @implements([aten.mm.default])
 def float8_mm(aten_op, args, kwargs=None):
-    assert isinstance(args[0], Float8Tensor) and isinstance(args[1], Float8Tensor)
     a = args[0]
     b = args[1]
+    if isinstance(b, AsyncCollectiveTensor):
```
drisspg, line 84: not sure how this is creeping in here

float8_experimental/float8_tensor.py
```diff
@@ -13,9 +13,43 @@
     to_fp8_saturated,
 )
 
+from torch.distributed._tensor import DTensor
+
 aten = torch.ops.aten
 
 
+def to_fp8_no_autograd(
```
drisspg, line 21: code_motion + handling DTensor

float8_experimental/float8_ops.py
wanchaol: this is a bit surprising, so this change fixed the AsyncCollectiveTensor issue?

float8_experimental/float8_tensor.py
```diff
@@ -13,9 +13,77 @@
...
+            x, DTensor
+        ), "Expected Float8 scale to be a DTensor if bits_fp8 is a DTensor"
+        bits_mesh = bits_fp8.device_mesh
+        bits_placements = bits_fp8.placements
+        local_bits = bits_fp8.to_local()
+        local_scale = x_scale.to_local()
+        inner_float8_tensor = Float8Tensor(
+            local_bits, local_scale, x.dtype, emulate=emulate
+        )
+        return DTensor.from_local(inner_float8_tensor, bits_mesh, bits_placements)
```
wanchaol, line 56: nit: let's use `run_check=False` option

float8_experimental/float8_ops.py
drisspg: Yeah, exactly, Im not totally sure yet how this collective made it into the dispatch  for Float8Tensor

float8_experimental/float8_tensor.py
```diff
@@ -13,9 +13,77 @@
...
+    def to_original_precision(grad):
+        if isinstance(grad, Float8Tensor):
+            return grad.to_original_precision()
+        else:
+            return grad
+
+    if isinstance(x, DTensor):
+        local_grad = x.to_local()
+        original_precision_grad = to_original_precision(local_grad)
+        return DTensor.from_local(original_precision_grad, x.device_mesh, x.placements)
```
wanchaol, line 82: ditto
```
DTensor.from_local(original_precision_grad, x.device_mesh, x.placements, run_check=False, shape=x.shape, stride=x.stride)

**Review by wanchaol — APPROVED** [review]
This looks pretty decent to me! have only one question about ACT and some minor suggestions

File: float8_experimental/float8_ops.py (hunk 1 above)
Comment (drisspg) [reply float8_experimental/float8_ops.py:84]: yeah tbh this is the only part I don't love I am having a hard time figure out where this is creeping in.. 

This dispatch should be DTensor(Float8Tensor()) -> aten.mm.defualt
.... okay as I am writing this I think I migh tnow


I think we have 
```
DTensor(Float8Tensor)), DTensor(AsyncTensor(Float8Tensor) -> aten.mm.defualt
```

The first layer of torch.dispatch goes through Dtensors table for both. Then we have unwrap
 
 ```
 Float8Tensor(), Async(Float8Tensor) -> aten.mm.default
 ```
 
 And the argparse chooses the leftmost subclass to dispatch too and thus we need this. I wonder if we can either handle this in DTensor or if there is another way?
 
 cc @bdhirsh  if this makes sense as well

Comment (albanD) [reply float8_experimental/float8_ops.py:84]: If your Float8Tensor class expects any other subclass to run before you, you can add this to your class:
```python
        if not all(issubclass(cls, t) for t in types):
            return NotImplemented
```

The NotImplemented will trigger the handler of the next type (just like it does for `__add__` in regular python)

drisspg: Thanks Alban this makes alot  of sense, just updated and everything appears to be working as expected!

PR #227, opened 2024-02-27
add option for using fused kernel

drisspg:


drisspg:
@y-sq  had a good suggestion to just use amax directly and not the scale to avoid launch of extra kernel: https://github.com/drisspg/driss_torch/pull/5

That will require moving the to_float8 constructor to take in an amax and not a scale, I will work on that tomorrow..

PR #228, opened 2024-02-27
Make code more dry and standardize on to_fp8_no_autograd

drisspg:
# Summary

Remove direct Float Constructor invocations inside torch.autograd functions and use the `to_fp8_no_autograd` function
The is also fixes a problem where the DTensor tests weren't in test_everything.sh which lead to the torch__dispatch being incorrect, also fixed now.

`./test/test_everything` is all green


wanchaol (approved):
lgtm!

PR #232, opened 2024-03-01
Add in pre-commit config and some more CI/CD

drisspg:
# Summary
Some quality of life changes

### Details:
- Use Ruff for linting, majority of file changes were are from this
- Add in pre-commit conifg so that user don't forget 😉
- Add both linting and testing github actions. NOTE: this is very non exhaustive, we don't currently have an gpu runners and even if we did they would need to be h100 or sm89 to significantly cover  the tests we have, but something is better than nothing

#### One auxiliary change 
- While doing this PR I noticed that the c10d ops have been updated to _c10d so registered this ops to float8 tensor class and test/test_dtensor works

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -27,9 +27,9 @@ jobs:
         python -m pip install --upgrade pip
         pip install -e .
         pip install -e .'[dev]'
-    - name: Lint with flake8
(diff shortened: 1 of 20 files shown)


wanchaol (approved):
always happy for lint/ci improvements!

PR #233, opened 2024-03-04
early return by check tensor already casted or not

wanchaol:
as titled, it turns out we don't need to install additional hooks base on our TP + FP8 design. The only thing we need to do here is to be able to turn off activation casting, so that we can put activation casting in the TP hooks

So just check if the tensor already been casted to fp8 or not, as in TP we would cast activation into the DTensor's Float8Colwise/Rowwise instead.


float8_experimental/float8_dynamic_linear.py
```diff
@@ -62,26 +62,26 @@ class Float8DynamicLinear(torch.nn.Linear):
...
         """
         super().__init__(**super_kwargs)
 
-        self.use_activation_hooks = use_activation_hooks
+        self.cast_activation = cast_activation
 
     def forward(self, x):
         # cast x to float8_e4m3fn if not using activation hooks
-        x_fp8 = x if self.use_activation_hooks else self.cast_to_float8_e4m3fn(x)
+        x_fp8 = self.cast_to_float8_e4m3fn(x) if self.cast_activation else x
```
awgu, line 76: If `self.cast_activation is False`, then what type is `x`? Would it be a `DTensor` whose local tensor is `Float8Tensor`?

I wonder if we need the bool at all. For example, the `cast_to_float8_e4m3fn` function could be idempotent in that if it is passed a `Float8Tensor`, then we return it early. The wrinkle here is that it could be a `DTensor` wrapping a `Float8Tensor`.

drisspg: I would imagine that its a DTensor(Float8Tensor) and that the only thing that changes is, do we call `Float8Tensor_constructor(DTensor(Tensor))` in the activation hook or actually in the forward

float8_experimental/float8_dynamic_linear.py
```diff
@@ -62,26 +62,26 @@ class Float8DynamicLinear(torch.nn.Linear):
...
 
         # cast w to float8_e4m3fn
         w_fp8 = self.cast_to_float8_e4m3fn(self.weight)
 
         y = torch.nn.functional.linear(x_fp8, w_fp8, self.bias)
 
         # Cast gradY to float8_e5m2 during backward if not using activation hooks
-        if not self.use_activation_hooks:
+        if self.cast_activation:
```
drisspg, line 84: I don't get why this the logic flipped here though

float8_experimental/float8_dynamic_linear.py
drisspg: Ohh so I guess, we cast activation is saying "DTensor is not going to handle this for you, so you need to do it?"

float8_experimental/float8_dynamic_linear.py
drisspg, line 84: DTensor also can handle this?

float8_experimental/float8_dynamic_linear.py
wanchaol: if cast activation is false, it would be `DTensor(torch.Tensor[fp32/16])`, we want to turn off activation casting is because we want this casting happen inside the TP preforward/forward hooks (as it needs to happen in certain order, i.e. after `from_local` and before `redistribute`), see changes in this PR https://github.com/pytorch-labs/float8_experimental/pull/234

float8_experimental/float8_dynamic_linear.py
wanchaol: yeah both activation casting before `F.linear` and after `F.linear` now happens in TP hooks directly in https://github.com/pytorch-labs/float8_experimental/pull/234

float8_experimental/float8_dynamic_linear.py
wanchaol: updated according to our discussions :)

float8_experimental/float8_dynamic_linear.py
```diff
@@ -30,63 +34,43 @@ def forward(
 
     @staticmethod
     def backward(ctx, gradY):
+        if tensor_already_casted_to_fp8(gradY):
```
wanchaol, line 37: for `cast_to_float8_e5m2_bw`, unfortunately I can do a forward check only and have to check the backward gradients to see if it's already casted, as the forward only have `y` but not `grad_y`

float8_experimental/float8_tensor.py
```diff
@@ -14,6 +14,19 @@
...
 
 
+def tensor_already_casted_to_fp8(tensor: torch.Tensor) -> bool:
+    """
+    Check if the tensor is already casted to fp8
+    """
+    if isinstance(tensor, Float8Tensor):
+        return True
+    elif isinstance(tensor, DTensor) and isinstance(tensor._local_tensor, Float8Tensor):
+        # TODO: shall we stick to public API and directly use tensor.to_local() here?
```
drisspg, line 24: I wonder if in general, subclasses composing with other subclasses should have a generic way to determine the nested types

drisspg (approved):
Awesome! I love the ratio of red lines to green!

I think the fixture in conftest.py for xfailing the grad hooks can also be removed

Thanks!

float8_experimental/float8_tensor.py
wanchaol: Yeah, I'll also need to think about how this should behave, this seems in general how we can get subclass hierarchy from some type info, simply trying out `type.mro` or `inspect.getmro` seems not working for me. Maybe need to check with @ezyang @bdhirsh to see if there's any better suggestions

float8_experimental/float8_linear.py
```diff
@@ -304,16 +304,15 @@ def forward(self, x):
...
-    def from_float(cls, mod, emulate: bool = False, use_activation_hooks: bool = False):
+    def from_float(cls, mod, emulate: bool = False):
         """
         Create an nn.Linear with fp8 compute from a regular nn.Linear
 
         Args:
             mod (torch.nn.Linear): nn.Linear to convert
             emulate (bool): whether to emulate fp8 matmul logic in float32
-            use_activation_hooks (bool): whether to use activation hooks instead of inlining the casting logic
+            cast_activation (bool): whether to use activation hooks instead of inlining the casting logic
```
awgu, line 314: nit: This looks like it should be removed.

drisspg: ahh good catch

drisspg: I think this was removed in the next pr

PR #234, opened 2024-03-04
Add support for Tensor Parallelism (DTensor) + Float8DynamicLinear

wanchaol:
This is based on the early return if casted changes, adding capability to perform tensor parallelism using PyTorch native way (DTensor). We introduce `Float8ColwiseParallel` and `Float8RowwiseParallel` style that's built on top of the Col/RowwiseParallel APIs from core, and extend that to make activation casting works with TP allgather/reduce_scatter.

The UX can be seen in `test_dtensor.py`

This also needs a PyTorch PR https://github.com/pytorch/pytorch/pull/120895


vkuzo:
API lgtm, what's remaining to make this landable?

wanchaol:
> API lgtm, what's remaining to make this landable?

I think this is landable now, both vanilla TP and "sequence parallel" works in the test

float8_experimental/float8_tensor_parallel.py
```diff
@@ -0,0 +1,103 @@
...
+from torch.distributed.tensor.parallel import ColwiseParallel, RowwiseParallel
+
+# subclass the ColwiseParallel and RowwiseParallel classes
+# to add the float8 support
+# The parameter sharding stays the same as the core
+# ColwiseParallel and RowwiseParallel, the only difference
+# here is that in input/output handling we do casting after
+# creating the DTensor.
+
+# NOTE: This only works and tested with the DynamicLinear
```
drisspg, line 13: Out of curiosity do we have a path for Delayed?

drisspg (approved):

float8_experimental/float8_tensor_parallel.py
wanchaol: I'm not quite familiar with delayed scaling yet, but from a quick look I think the approach we are building with dynamic scaling might be working with delayed scaling, might need some more changes and testing :)

PR #235, opened 2024-03-07
Adds utilities for AMD fp8 dtype support, follow up PR to add option to the configs

drisspg:
# Summary

AMD GPUS support a different fp8 dtype compared to nvidia. These dtypes were added to PyTorch and we update Float8Tensor construction to use the format dependent on the arch.

For a detailed summary see: https://github.com/openxla/stablehlo/blob/main/rfcs/20230321-fp8_fnuz.md

### DIFF
--- a/float8_experimental/float8_dynamic_linear.py
+++ b/float8_experimental/float8_dynamic_linear.py
@@ -88,7 +88,7 @@ class Float8DynamicLinear(torch.nn.Linear):
                 "bias": False,
             }
             new_mod = cls(**super_kwargs)
-        new_mod.forward_config = ScaledMMConfig(emulate, True if not emulate else False)
+        new_mod.forward_config = ScaledMMConfig(emulate, not bool(emulate))
         new_mod.backward_config = ScaledMMConfig(emulate, False)
         if config.enable_fsdp_fp8_all_gather:
             new_mod.weight = nn.Parameter(
--- a/float8_experimental/float8_utils.py
+++ b/float8_experimental/float8_utils.py
@@ -12,19 +12,17 @@ import torch.distributed as dist
 # Helpful visualizer for debugging (only supports fp32):
 # https://www.h-schmidt.net/FloatConverter/IEEE754.html
 
-# define the e4m3/e5m2 constants
-E4M3_MAX_POS = torch.finfo(torch.float8_e4m3fn).max
-E4M3_FNUZ_MAX_POS = torch.finfo(torch.float8_e4m3fnuz).max
-E5M2_MAX_POS = torch.finfo(torch.float8_e5m2).max
-E5M2_FNUZ_MAX_POS = torch.finfo(torch.float8_e5m2fnuz).max
-
-FP16_MAX_POS = torch.finfo(torch.float16).max
-
 # avoid division by zero when calculating scale
 # TODO: align this value with NVIDIA's assumptions (current value is a guess)
 EPS = 1e-12
 
 IS_AMD = torch.cuda.is_available() and torch.version.hip is not None
+FP8_TYPES = {
+    torch.float8_e4m3fn,
+    torch.float8_e5m2,
+    torch.float8_e4m3fnuz,
+    torch.float8_e5m2fnuz,
+}
 
 
 @torch.no_grad()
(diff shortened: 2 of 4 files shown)


float8_experimental/float8_dynamic_linear.py
```diff
@@ -30,18 +30,17 @@ def forward(
 
     @staticmethod
     def backward(ctx, gradY):
-        gradY_scale = tensor_to_scale(gradY, torch.float8_e5m2)
-        fp8_tensor = to_fp8_no_autograd(
-            gradY, gradY_scale, torch.float8_e5m2, ctx.emulate
-        )
+        fp8_dtype = torch.float8_e5m2fnuz if IS_AMD else torch.float8_e5m2
```
gchanan, line 33: should this be configurable in the forward with a reasonable default?

malfet: I wonder if it would be better to have `torch.backends.[cuda|hip|mps].supports_dtype(XYZ)` API, because I assume XPUs would probably use `fnuz` flavor, but say ARM would be using `e5m2` flavor

vkuzo: I think this should be configurable instead of depending on env, numerics should be as predictable as possible. It's also valuable to emulate numerics without having the hardware, for debugging.

drisspg: Oaky, so it sounds like we want this to be defined at module construction.  It is up to the constructor of the module to ensure that _scaled_mm will work with  their module.

I do think that Nikitas backend dtype helper could a useful pytorch feature, not sure if that should live here though

drisspg:
The non emulated version is failing for me on on an mi300 machine using this version of pytorch:
```Shell
pytorch-triton-rocm==3.0.0+0a22a91d04
torch==2.3.0.dev20240308+rocm6.0
```

``` Shell
FAILED test/test_base.py::TestFloat8Linear::test_linear_nobias[True-LinearType.DYNAMIC-x_shape0-False] - AssertionError: -2.7592885494232178 is too low
FAILED test/test_base.py::TestFloat8Linear::test_linear_nobias[True-LinearType.DYNAMIC-x_shape1-False] - AssertionError: -3.372152805328369 is too low
FAILED test/test_base.py::TestFloat8Linear::test_linear_nobias[True-LinearType.DYNAMIC-x_shape2-False] - AssertionError: -2.8420748710632324 is too low
FAILED test/test_base.py::TestFloat8Linear::test_linear_nobias[False-LinearType.DELAYED-x_shape0-False] - AssertionError: -2.7584447860717773 is too low
FAILED test/test_base.py::TestFloat8Linear::test_linear_nobias[False-LinearType.DELAYED-x_shape1-False] - AssertionError: -2.946033239364624 is too low
FAILED test/test_base.py::TestFloat8Linear::test_linear_nobias[False-LinearType.DELAYED-x_shape2-False] - AssertionError: -2.756319999694824 is too low
FAILED test/test_base.py::TestFloat8Linear::test_linear_nobias[False-LinearType.DYNAMIC-x_shape0-False] - AssertionError: -3.377957820892334 is too low
FAILED test/test_base.py::TestFloat8Linear::test_linear_nobias[False-LinearType.DYNAMIC-x_shape1-False] - AssertionError: -3.0644452571868896 is too low
FAILED test/test_base.py::TestFloat8Linear::test_linear_nobias[False-LinearType.DYNAMIC-x_shape2-False] - AssertionError: -3.091813564300537 is too low
```

alugorey:
@drisspg Is this mergeable? The errors above should have been fixed with this: https://github.com/pytorch/pytorch/pull/125921

drisspg:
@alugorey I was actually thinking that something similiar would be landed here: https://github.com/pytorch-labs/float8_experimental/pull/248

alugorey:
> @alugorey I was actually thinking that something similiar would be landed here: #248

Ah okay, I was working on top of this PR. Do you want me to pull your changes into my PR and we abandon this one? Or do you want me to point my PR to this branch?

drisspg:
@alugorey ahh no worries, let met rebase and whip this PR back into shape

float8_experimental/float8_utils.py
```diff
@@ -14,22 +14,40 @@
...
     scale = torch.empty_like(amax, dtype=torch.float32)
     if float8_dtype == torch.float8_e4m3fn:
         res = E4M3_MAX_POS / torch.clamp(amax, min=EPS)
-    else:  # e5m2
+    elif float8_dtype == torch.float8_e4m3fnuz:
+        res = E4M3_FNUZ_MAX_POS / torch.clamp(amax, min=EPS)
+    elif float8_dtype == torch.float8_e5m2:
         res = E5M2_MAX_POS / torch.clamp(amax, min=EPS)
+    elif float8_dtype == torch.float8_e5m2fnuz:
+        res = E5M2_FNUZ_MAX_POS / torch.clamp(amax, min=EPS)
```
malfet, line 48: Plz avoid code duplication
```suggestion
    elif float8_dtype in [torch.float8_e4m3fnuz, torch.float8_e5m2, torch.float8_e5m2fnuz]:
        res = torch.finfo(dtype).max / torch.clamp(amax, min=EPS)
```

malfet (approved):

float8_experimental/float8_utils.py
malfet: Actually, you don't even need ifs there, just assert that float8_dtype is indeed the one
```python
  assert float8_dtype.itemsize == 1 and float8_dtype.is_floating_point
```

PR #241, opened 2024-03-20
Remove fairscale tp/sp

drisspg:
# Summary
Now that we have DTensor support remove existing fairscale implementations of TP/SP

As well I fixed the test_sam.py not being able to run and an issue with tmp file overriding Pythons path lookup


vkuzo (approved):

test/test_sam.py
```diff
@@ -55,7 +60,12 @@ def test_encoder_fw_bw(self, data_dtype):
         ref_name_to_grad = {
             name: param.grad for name, param in encoder_ref.named_parameters()
         }
-        sqnr_threshold = 1.0 if data_dtype == torch.float16 else -4
+
+        # Delayed scaling has less performant numerics
+        fudge_factor = 7.0 if linear_type == Float8Linear else 1.0
+        sqnr_threshold = -1.0 if data_dtype == torch.float16 else -4
```
vkuzo, line 66: would be good for a threshold value to be positive

PR #248, opened 2024-04-01
Changes on top of upstream to get rid of type errors

alugorey:
Fixes the class of failed unit tests  on rocm in test_base.py that fail the internal assertion `Cannot convert ScalarType Float8_e4m3fn to hipDataType.`

Note: We are aware of the outstanding numerical issues and are looking into it internally.

### DIFF
--- a/float8_experimental/config.py
+++ b/float8_experimental/config.py
@@ -19,3 +19,7 @@ enable_pre_and_post_forward = True
 # implements pre/post-all-gather methods to do fp8 all-gather with FSDP2.
 # Only dynamic scaling is supported for now.
 enable_fsdp_fp8_all_gather = False
+
+# If True, use 'fnuz' float8 types for calculations.
+# Currently, ROCm only supports fnuz variants.
+use_fnuz_dtype = False
--- a/float8_experimental/float8_utils.py
+++ b/float8_experimental/float8_utils.py
@@ -6,10 +6,10 @@
 
 from typing import Literal, Tuple
 
+import float8_experimental.config as config
+
 import torch
 import torch.distributed as dist
-
-import float8_experimental.config as config
 
 # Helpful visualizer for debugging (only supports fp32):
 # https://www.h-schmidt.net/FloatConverter/IEEE754.html
--- a/float8_experimental/float8_utils.py
+++ b/float8_experimental/float8_utils.py
@@ -9,6 +9,8 @@ from typing import Literal, Tuple
 import torch
 import torch.distributed as dist
 
+import float8_experimental.config as config
+
 # Helpful visualizer for debugging (only supports fp32):
 # https://www.h-schmidt.net/FloatConverter/IEEE754.html
 
@@ -16,13 +18,18 @@ import torch.distributed as dist
 # TODO: align this value with NVIDIA's assumptions (current value is a guess)
 EPS = 1e-12
 
-IS_AMD = torch.cuda.is_available() and torch.version.hip is not None
+IS_ROCM = torch.cuda.is_available() and torch.version.hip is not None
 FP8_TYPES = {
     torch.float8_e4m3fn,
     torch.float8_e5m2,
     torch.float8_e4m3fnuz,
     torch.float8_e5m2fnuz,
 }
+
+
+# User defined type for using the individual F8 type based on config
+e4m3_dtype = torch.float8_e4m3fn if not config.use_fnuz_dtype else torch.float8_e4m3fnuz
+e5m2_dtype = torch.float8_e5m2 if not config.use_fnuz_dtype else torch.float8_e5m2fnuz
 
 
 @torch.no_grad()
@@ -148,7 +155,7 @@ def compute_error(x: torch.Tensor, y: torch.Tensor):
 
 
 def fp8_tensor_statistics(
-    tensor: torch.Tensor, float8_dtype=torch.float8_e4m3fn
+    tensor: torch.Tensor, float8_dtype=e4m3_dtype
 ) -> Tuple[int, ...]:
     """Calculate FP8 tensor stats
 
--- a/test/test_base.py
+++ b/test/test_base.py
@@ -30,11 +30,11 @@ from float8_experimental.float8_tensor import (
 )
 from float8_experimental.float8_utils import (
     compute_error,
+    e4m3_dtype,
+    e5m2_dtype,
     fp8_tensor_statistics,
     FP8_TYPES,
     tensor_to_scale,
-    e4m3_dtype,
-    e5m2_dtype,
 )
 
 random.seed(0)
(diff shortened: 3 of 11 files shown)


alugorey:
@drisspg :)

drisspg:
Awesome! I will take a look at this tomorrow

float8_experimental/float8_utils.py
```diff
@@ -28,12 +28,25 @@
 IS_AMD = torch.cuda.is_available() and torch.version.hip is not None
 
 
+# Helper functions to get individual F8 types based on backend architecture
```
vkuzo, line 31: Would it be possible to put this into configuration instead of setting it dynamically?  It can be unexpected for numerics to change based on the environment.  It would also be good to support numerical emulation of all of these types regardless of whether the user's machine supports a float8 matmul.

alugorey: I'm afraid I don't understand your question. These helper functions are simply intended to grab the "right" version of the prebuilt torch F8 types. Could you elaborate on the change you'd like to see?

vkuzo: Sure, it's just making the dtype flavors encoded in configuration instead of environment dependent.  Having this in configuration would make it easier to debug numerics without having the target hardware.

```
# float8 dtypes have a default which can be changed explicitly
config = ...
config.float8_flavors = 'nuz'
do_float8_things(..., config)
```

versus

```
# float8 dtypes magically change based on the environment
do_float8_things(...)
```

That said, my comment is not high pri, feel free to land and we can adjust this later if it becomes important.

test/test_base.py
```diff
@@ -350,7 +374,7 @@ def test_scaled_mm_vs_emulated(self, base_dtype):
 
 
 class TestNumerics:
-    @pytest.mark.parametrize("float8_dtype", [torch.float8_e4m3fn, torch.float8_e5m2])
+    @pytest.mark.parametrize("float8_dtype", [fp8_e4m3_t(), fp8_e5m2_t()])
```
vkuzo, line 377: I would recommend testing all cases on all hardware types instead.  For things not requiring a matmul, it should just work.  For things requiring a matmul, we have an emulation mode to at least help approximate it.

test/test_base.py
```diff
@@ -47,7 +51,10 @@ class TestFloat8Tensor(unittest.TestCase):
     def test_preserves_dtype(self) -> None:
         # hp means high precision, lp means low precision
         hp_dtypes = (torch.float32, torch.float16, torch.bfloat16)
-        lp_dtypes = (torch.float8_e4m3fn, torch.float8_e5m2)
+        fp8_dtypes = (
+            FP8Dtypes()
```
vkuzo, line 55: all dtypes would be nice, there should not be anything in `Float8Tensor` which is hardware dependent

test/test_base.py
```diff
@@ -259,8 +274,13 @@ def test_type_cast(
...
         )
 
+
+        fp8_dtypes = (
+            FP8Dtypes()
+        )
+
         m = nn.Linear(32, 16, device="cuda", dtype=linear_dtype)
-        m = get_float8_linear(linear_type, m, emulate, False)
+        m = get_float8_linear(linear_type, m, emulate, False, fp8_dtypes)
```
vkuzo, line 283: you can enable emulation here if your hardware doesn't support the dtype under test

float8_experimental/float8_utils.py
alugorey: @vkuzo Sorry, got wrapped up in other work recently and just circled back to this. Okay, I will add an option in float8_experimental/float8_experimental/config.py and instead of checking the backend architecture, the code will check against this user-settable config variable

alugorey:
@vkuzo Ready for another review. Also, wanted to ask if there was an ETA or roadmap for when this functionality would be pulled into pytorch proper?

vkuzo:
so sorry I am on holiday right now, will take a look late next week when I return, unless @drisspg wants to get to it sooner

drisspg:
Yeah will review tomorrow

drisspg (commented):
What was the output of test_everything.sh?

drisspg (commented):
What was the output of test_everything.sh?

alugorey:
@drisspg 
[test_everything.log](https://github.com/pytorch-labs/float8_experimental/files/15167634/test_everything.log)

Fails in test_compile. However, I was aware of this failure and found that this is unrelated to my changes, rather, an issue with torch.compile on ROCm. This failure was next on my TODO's to address. I will upload a follow up PR.

float8_experimental/config.py
```diff
@@ -19,3 +19,8 @@
 # TODO(before land): add test coverage for both cases
 # dynamic_use_activation_hooks = True
 # dynamic_use_activation_hooks = False
+
+# If True, use 'fnuz' float8 types for calculations. If the backend
+# hardware does not support a particular type, the emulated implementation
+# of the dtype will be used. Currently, ROCm only supports the fnuz variants.
+use_fnuz_dtype = True
```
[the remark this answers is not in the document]
alugorey: ah, yes, that's an oversight due to being easier for me to test. will change.

test/test_dtensor.py
```diff
@@ -128,7 +128,7 @@ def test_dtensor_fp8_autograd(mesh: DeviceMesh, size=16):
     )
 
     out = torch.nn.functional.linear(dist_x_fp8, dist_weight_fp8)
-    out = NoopFwToFloat8E5M2Bw.apply(out, False)
+    out = NoopFwToFloat8E5M2Bw.apply(out, False, fp8_e5m2_t())
```
vkuzo, line 131: is the new last arg here expected?

alugorey: Oversight, fixed in latest commit

drisspg:
just an update here, the base PR should landed yesterday

alugorey:
@vkuzo @drisspg 
Revised/rebased drop of AMD support after the `amd-support` branch was merged to main

float8_experimental/config.py
```diff
@@ -19,3 +19,8 @@
 # implements pre/post-all-gather methods to do fp8 all-gather with FSDP2.
 # Only dynamic scaling is supported for now.
 enable_fsdp_fp8_all_gather = False
+
+# If True, use 'fnuz' float8 types for calculations. If the backend
+# hardware does not support a particular dtype, the emulated implementation
```
vkuzo, line 24: nit: currently the user is responsible for toggling the emulation setting, we don't do that automatically

vkuzo (approved):
looks great! thanks for helping!

float8_experimental/config.py
alugorey: I thought per previous comments that we wanted to go the emulated route if the backend hardware didn't support the type? in general, the user could force to emulate, but i was under the impression that cases where a dtype was being used that wasn't supported on the underlying hardware, that we wanted to go the emulated route?

vkuzo: yep, that is correct!  This is just currently done by the user explicitly, and there is no support do handle that automatically.

alugorey: oh, you're saying the comment is wrong

alugorey: fixed!

alugorey:
@drisspg fixed lint. But not sure what the ufmt errors are about.

alugorey:
@drisspg Looks like you beat me to it. Still failing on 2 files though. is there documentation on what ufmt expects? some of those changes seem purely cosmetic

drisspg:
The reason for this formatting is due to some internal nodes on code styling. TBH
I just run `ufmt format .` and dont care about the actual format.

I have also have some 'pre-commit' hooks for the repo that should be set and forget. There does seem to be 2 more lint fixes. 

There seems to be 1 real test_failure:

```
./test/test_everything.sh: line 5: rocm-smi: command not found
```

I found this patch passed test_everything:

https://gist.github.com/drisspg/47a29d6bf3fcca2a2c48d09b74c564aa

Issue #238, opened 2024-03-17
torch.inference_mode switches`aten.linear.default, this is not supported`


michaelfeil:
First, congrats for the repo - looks great

I discovered that switching between `torch.no_grad` and `torch.inference_mode` leads to a switch to `aten.linear.default`. Feel free to use this feedback - its likely expected.

```
# dependencies
'2.4.0.dev20240316+cu121'
float8_experimental commit 88e9e507c56e59c5f17edf513ecbf621b46fc67d
```

```python
from transformers import AutoModel, AutoTokenizer, PreTrainedModel
import torch

from float8_experimental.float8_linear_utils import (
    swap_linear_with_float8_linear,
)
from float8_experimental.float8_dynamic_linear import Float8DynamicLinear

"""
pip3 install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu121 --upgrade
"""

model_name = "michaelfeil/bge-small-en-v1.5"
model: PreTrainedModel = AutoModel.from_pretrained(model_name)
model_orig: PreTrainedModel = AutoModel.from_pretrained(model_name).cuda()

tokenizer = AutoTokenizer.from_pretrained(model_name)
tokens = tokenizer.batch_encode_plus(["hi","there is no"], return_tensors="pt", padding=True, truncation=True)
tokens = {k: v.cuda() for k,v in tokens.items()}

model = model.to("cpu")
swap_linear_with_float8_linear(model, Float8DynamicLinear)
model = model.to("cuda")

with torch.no_grad():
    out1 = model_orig.forward(**tokens)["last_hidden_state"]
    out2 = model.forward(**tokens)["last_hidden_state"]

print(out1 - out2, out1, out2, (out1-out2).mean())

with torch.inference_mode():
   # breaks with NotImplementedError: attempting to run aten.linear.default, this is not supported
    out2 = model.forward(**tokens)["last_hidden_state"]
```

drisspg:
Hey, thanks for opening this issue! Let me take a look at this

tgaddair:
Hey @drisspg any updates on this? Fixing this would be a enabler for a lot of serving use cases.

drisspg:
Hey, thanks for repinging! I just added a one liner that unblocked the repro, still need to tidy some things up.

I do think that this will allow you to experiment further with fp8 for inference, however I do not think that this is the ideal usage. FP8 will benefit from  what is usual done in other int8 quantization techniques. 

[...]

PR #251, opened 2024-04-10
Need to figure out the correct error propagation and add tests

drisspg:
# Summary
This enables logic that work with torch.no.grad to work with inference mode:  #238

### DIFF
--- a/float8_experimental/float8_tensor.py
+++ b/float8_experimental/float8_tensor.py
@@ -325,6 +325,8 @@ class Float8Tensor(torch.Tensor):
 
         if func in FLOAT8_OPS_TABLE:
             return FLOAT8_OPS_TABLE[func](func, args, kwargs)
+        else:
+            return func.decompose(*args, *kwargs)
         raise NotImplementedError(f"attempting to run {func}, this is not supported")
 
     # Do not force the Float8Tensor type on the returned tensor


albanD:
Huh. I don't know off the top of my head what the decompose method is doing... @zou3519 might know better?

float8_experimental/float8_tensor.py
```diff
@@ -325,6 +325,8 @@ def allowed_subclasses(type):
 
         if func in FLOAT8_OPS_TABLE:
             return FLOAT8_OPS_TABLE[func](func, args, kwargs)
+        else:
+            return func.decompose(*args, *kwargs)
```
zou3519, line 329: I'm not sure this is what you want, it directly invokes the CompositeImplicitAutograd implementation

albanD: I think this is exactly what they want :D 
What happens if there is no CompositeImplicitAutograd impl?

zou3519: it returns NotImplemented

drisspg: yeah I figure that it we can noop ( execept not implemented) slide our way down by trying to decompose the op 
- Either it wont be decomposable, and return not implemented, cool same place as after this
- or it is decomposable:
	- Run supported op, Great!
	- Not supported and ultimately end back up with a not implemented

PR #253, opened 2024-04-25
Enable restricted split + cat in order to enable SP

drisspg:
# Summary
This comes from needing to support sequence parallelism in torchtitan

### DIFF
--- a/float8_experimental/float8_ops.py
+++ b/float8_experimental/float8_ops.py
@@ -3,7 +3,7 @@
 #
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
-from typing import Any, Dict
+from typing import Any, Dict, Tuple
 
 import torch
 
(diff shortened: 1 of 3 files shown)


float8_experimental/float8_ops.py
```diff
@@ -50,7 +50,38 @@ def float8_desugar_op(aten_op, args, kwargs=None):
...
+    def make_float8(data):
+        return Float8Tensor(
+            data, args[0]._scale, args[0]._orig_dtype, args[0]._mm_config
+        )
+
+    out = map(make_float8, new_data_tensors)
+    return list(out)
+
+
+# Errors cant `cat_cuda float8 e4m3fn`
```
wanchaol, line 66: oh so this means that the `torch.cat` can't apply to dtype e4m3fn?

Normally I feel this is something that we can just make our cuda kernel to support concatting tensors with the same dtype, but not sure if there're further complications there for fp8 dtype.

But if the job we want to do is to simply concatting the fp8 tensors together, one simpler way to can do:

we can just try to do `fp8_inner_tensor.view(torch.uint8)`, perform `torch.cat`, then after the cat operation, we do `fp8_catted_tensor.view(torch.float8_e4m3fn`, I wonder if this would unblock?

wanchaol (approved):
sgtm! thanks for supporting this!

PR #255, opened 2024-05-03
make the backward of differentiable float8 casts pass gradient as is

vkuzo:
Summary:

Behavior before:
* high precision to float8 in fw, float8 to high precision in bw
* float8 to high precision in fw, high precision to float8 in bw if grad is a Float8Tensor, pass gradient unchanged otherwise

Behavior after:
* high precision to float8 in fw, pass gradient unchanged in bw
* float8 to high precision in fw, pass gradient unchanged in bw

Motivation for the new state:
1. we want gradients to be in high precision unless specified otherwise by the float8 recipe, and the logic to specify grad casting to float8 before the matmul is better implemented elsewhere
2. there is actually no logic change in this diff as the backward casts were not getting hit from existing code, this diff just makes the intended behavior clearer

Test Plan:

```
./test/test_everything.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/float8_experimental/float8_tensor.py
+++ b/float8_experimental/float8_tensor.py
@@ -112,43 +112,12 @@ def to_fp8_no_autograd(
     return Float8Tensor(bits_fp8, x_scale, x.dtype, mm_config=mm_config)
 
 
-def from_fp8_no_autograd(x: torch.Tensor) -> torch.Tensor:
-    """Convert a tensor from float8 without autograd
-
-    This function will handle 3 cases:
-        1. If the tensor is a DTensor, it will convert the inner tensor to the original precision
-        2. If the tensor is a Float8Tensor, it will convert the tensor to the original precision
-        3. If the tensor is a regular tensor, it will pass through this tensor
-
-    Args:
-        x: the tensor to convert
-    """
-
-    def to_original_precision(grad):
-        if isinstance(grad, Float8Tensor):
-            return grad.to_original_precision()
-        else:
-            return grad
-
-    if isinstance(x, DTensor):
-        local_grad = x.to_local()
-        original_precision_grad = to_original_precision(local_grad)
-        return DTensor.from_local(
-            original_precision_grad,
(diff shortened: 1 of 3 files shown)


vkuzo:
I noticed this logic was funky when copy-pastaing some of this code for MX prototyping, fixing here to clarify intent.

float8_experimental/float8_tensor.py
```diff
@@ -112,43 +112,12 @@ def to_fp8_no_autograd(
...
-        x: the tensor to convert
-    """
-
-    def to_original_precision(grad):
-        if isinstance(grad, Float8Tensor):
-            return grad.to_original_precision()
-        else:
-            return grad
-
-    if isinstance(x, DTensor):
```
wanchaol, line 133: I recalled the main reason we had that special handling is that torch.compile (specifically `FakeTensor` rule) can't take nested subclass rule automatically. If we remove this wondering where we would put such a logic? 

cc @drisspg @bdhirsh

vkuzo: if I understand the old state correctly:
1. `grad` is never a `Float8Tensor` in practice, because we always have the matmul output high precision
2. things already work without nested subclasses, which is why removing this function is fine

But, would be great to clarify ^

wanchaol: Oh I checked the current workflow, we call:
1. cast_to_float8_e4m3fn
2. cast_to_float8_e5m2_bw
and it seems none of these calls would call into this from_fp8_no_autograd

so this should be fine

wanchaol: actually it looks like I'm wrong, the `cast_to_float8_e4m3fn`'s backward would call this, but maybe it's a no-op even as of current state, as the output of the torch.scaled_mm in backward is fp32?

wanchaol (approved):

drisspg (approved):

PR #266, opened 2024-05-21
[FSDP2] precompute scale after optimizer.step for dynamic scaling

weifengpy:
Goal: improve float8 all-gather perf in FSDP2 by precomputing scales for all float8 params with a single all-reduce

updated README for API usage: call `precompute_float8_scale_for_fsdp` inside the training loop after optimizer step

```
from float8_experimental.fsdp_utils import precompute_float8_amax_for_fsdp
# inside the training loop
model(input).sum().backward()
optim.step()
precompute_float8_dynamic_scale_for_fsdp(model)
```

unit test `pytest -s test/test_fsdp2/test_fsdp2_eager.py -k test_transformer_parity_dynamic`

**FSDP pre-forward**: shortend from 3ms to 1.8ms because of doing 1 all-reduce instead N small all-reduces
Screenshot 2024-05-30 at 12 38 24 AM

Screenshot 2024-05-30 at 12 48 14 AM

**Pre-computing amax**: shortened from 5ms to 1.7ms, by switching from `torch._foreach_abs` + `torch.max(a)` to `torch._foreach_norm(weights, ord=math.inf)`

Screenshot 2024-05-30 at 12 50 17 AM

Screenshot 2024-05-30 at 12 49 54 AM


float8_experimental/float8_dynamic_linear.py
```diff
@@ -144,13 +144,19 @@ def __new__(cls, tensor: torch.Tensor, mm_config: ScaledMMConfig):
...
+            # pin_memory=tensor.is_pinned(),
+            pin_memory=False,
             requires_grad=tensor.requires_grad,
         )
 
     def __init__(self, tensor: torch.Tensor, mm_config: ScaledMMConfig):
         self._tensor = tensor
         self._mm_config = mm_config
+        # Optional cache for pre-computed fp8 data/scale
+        self._fp8_data: Optional[torch.Tensor] = None
```
bdhirsh, line 157: One major requirement for tensor subclasses that I don't think is respected here: `__tensor_flatten__` and `__tensor_unflatten__` must properly convey every inner tensor on the subclass.

So when we call `__tensor_flatten__` on this subclass, if either of `_fp8_data/scale/amax` are set to valid tensors, they need to be returned there (and similarly `__tensor_unflatten__` needs to handle them as extra args)

weifengpy: thanks for pointing this out! This saves me a lot of debugging time. I can give it a try by including `_fp8_data/scale/amax` in `__tensor_flatten__` and `__tensor_unflatten__`

float8_experimental/float8_linear_utils.py
```diff
@@ -322,3 +328,34 @@ def inner_func():
...
+        if isinstance(m, Float8DynamicLinear)
+        and isinstance(m.weight, DTensor)
+        and isinstance(m.weight._local_tensor, WeightWithDynamicFloat8CastTensor)
+    ]
+    weights: List[DTensor] = [float8_linear.weight for float8_linear in float8_linears]
+
+    def compute_amaxes(weights: List[DTensor]):
+        max_weights = torch._foreach_norm(weights, ord=math.inf)
+        amax_tensor = torch.vstack(max_weights)
+        amax_tensor = torch.clamp(amax_tensor, EPS)  # R
```
weifengpy, line 350: `torch.clamp` calls `all_reduce`. since we performed `torch.clamp` here, I avoided calling it again in `amax_to_scale(clamp_amax=False)`

float8_experimental/float8_dynamic_linear.py
```diff
@@ -190,9 +191,20 @@ def __repr__(self):
         return f"WeightWithDynamicFloat8CastTensor(tensor={self._tensor}, mm_config={self._mm_config})"
 
     def fsdp_pre_all_gather(self, mesh):
```
weifengpy, line 193: if `_pre_computed_amax`, we skip `tensor_to_amax` and directly do `amax_to_scale`

float8_experimental/float8_dynamic_linear.py
weifengpy: `torch.compile` works after patching https://github.com/pytorch/pytorch/pull/127431 
will compare traces in 2nd PR

float8_experimental/float8_linear_utils.py
awgu: So you are relying on `torch.clamp` to run the all-reduce implicitly from changing sharding from partial to replicate?

If this fragments the code, could we just all-reduce the amax tensor and then leave the clamp to `amax_to_scale`? I agree the current way is faster since we are doing one clamp for all amaxes, but in case float8 folks are not happy with this fragmentation, this seems like another way.

awgu (commented):
This seems reasonable to me! I want to check with float8 folks on the `amax_to_scale` change.

float8_experimental/float8_linear_utils.py
weifengpy: thanks for the suggestions. I can collect feedback from float8 folks if they have a preference

vkuzo:
nice! Can we include the intended user API in the PR summary?

float8_experimental/float8_dynamic_linear.py
```diff
@@ -151,6 +151,7 @@ def __new__(cls, tensor: torch.Tensor, mm_config: ScaledMMConfig):
     def __init__(self, tensor: torch.Tensor, mm_config: ScaledMMConfig):
         self._tensor = tensor
         self._mm_config = mm_config
+        self._pre_computed_amax = None
```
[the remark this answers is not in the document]
drisspg: +1 on adding to flatten/unflatten

float8_experimental/float8_linear_utils.py
drisspg: agreed

float8_experimental/float8_utils.py
```diff
@@ -27,17 +27,24 @@
...
     Args:
         amax: The amax value of the tensor.
         float8_dtype: The float8 dtype.
         orig_dtype: The original dtype of the tensor.
+        clamp_amax: default is True. False for FSDP fp8 all-gather since FSDP applied `torch.clamp` during pre-compute after optimizer.step
     """
     scale = torch.empty_like(amax, dtype=torch.float32)
     if float8_dtype in FP8_TYPES:
-        res = torch.finfo(float8_dtype).max / torch.clamp(amax, min=EPS)
+        if clamp_amax:
```
drisspg, line 44: nit: I think if you have this on a seperate line
`amax = clamp(amax, eps) if clamp_amax else amax`

makes the logic a lil easier to follow

float8_experimental/float8_linear_utils.py
```diff
@@ -322,3 +328,34 @@ def inner_func():
     for child in fp8_layers:
         # Set a flag to signal amaxes/scales are ready
         child.amax_and_scale_synced = True
+
+
+def precompute_float8_amax(module: nn.Module) -> None:
```
bdhirsh, line 333: @weifengpy do you plan / want to use compile on this, and are there any gaps around here that you think would be good to prioritize on the compile side?

This is mostly just me remembering @awgu mention a while ago that he thought compile added noticeable runtime overhead, and I can't remember if it was for this specific case. If it is, and we think compiling this code would be useful, I can prioritize looking into the runtime overhead.

weifengpy: Hi @bdhirsh, I plan to polish and land this PR without compile next week to conclude H1. most importantly add `_pre_computed_amax` to flatten/unflatten

Reducing runtime overhead from `torch.compile` is still meaningful since we want torch.compile(fp8 casting) in FSDP2 pre-forward hooks. would it be helpful if I work on a mini repro with profiler traces? Want to unblock you in the short-term

bdhirsh: If you have a mini repro showing bad runtime overheads with compile, that would be great!

weifengpy: Hi @bdhirsh , I have created a repro https://github.com/pytorch/pytorch/issues/129457 . I highlighted extra cpu overhead and gpu time for torch.compile(mode="reduce-overhead")

float8_experimental/float8_dynamic_utils.py
```diff
@@ -130,20 +143,38 @@ def unwrap(t):
         )
 
     def __tensor_flatten__(self):
-        return ["_tensor"], self._mm_config
+        if self._pre_computed_amax:
```
weifengpy, line 146: use if-else to handle optional tensor `_pre_computed_amax` otherwise dynamo will complain about `None`

float8_experimental/float8_dynamic_utils.py
```diff
@@ -9,10 +9,7 @@
 
 from typing import Any, Optional, Tuple
 
-import float8_experimental.config as config
-
 import torch
-import torch.nn as nn
```
weifengpy, line 15: pre-commit hook trigger linter, and cleaned unused import

float8_experimental/float8_linear_utils.py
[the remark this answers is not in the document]
weifengpy: moving to `fsdp_utils.py` according to PR https://github.com/pytorch-labs/float8_experimental/pull/310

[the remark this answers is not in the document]
weifengpy: indicating fsdp by renaming to `precompute_float8_amax_for_fsdp`

weifengpy:
> nice! Can we include the intended user API in the PR summary?

added example API usage in PR summary

float8_experimental/float8_linear_utils.py
```diff
@@ -322,3 +328,34 @@ def inner_func():
...
+        m
+        for m in module.modules()
+        if isinstance(m, Float8DynamicLinear)
+        and isinstance(m.weight, DTensor)
+        and isinstance(m.weight._local_tensor, WeightWithDynamicFloat8CastTensor)
+    ]
+    weights: List[DTensor] = [float8_linear.weight for float8_linear in float8_linears]
+
+    def compute_amaxes(weights: List[DTensor]):
+        max_weights = torch._foreach_norm(weights, ord=math.inf)
```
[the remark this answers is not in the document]
weifengpy: done

float8_experimental/float8_dynamic_linear.py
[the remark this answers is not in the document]
weifengpy: done

test/test_fsdp2/test_fsdp2_common.py
```diff
@@ -1,12 +1,12 @@
...
-from typing import List, Type
+from typing import List
 
 import float8_experimental.config as config
 
 import torch
 import torch.distributed as dist
 import torch.nn as nn
-from float8_experimental.float8_linear import Float8Linear
+from float8_experimental.fsdp_utils import precompute_float8_amax_for_fsdp
```
awgu, line 9: I might be very confused, but where is this `fsdp_utils` file?
Screenshot 2024-07-09 at 7 20 22 PM

and a nit: maybe should be consistent in whether `pre_compute` vs. `precompute` across the PR
(either way is fine to me, `precompute` maybe since it is more concise)

weifengpy: > I might be very confused, but where is this fsdp_utils file?

just added `float8_experimental/fsdp_utils.py`

> aybe should be consistent in whether pre_compute vs. precompute across the PR

good suggestion. I will use `precompute` consistently

float8_experimental/fsdp_utils.py
```diff
@@ -0,0 +1,58 @@
...
+import torch
+import torch.nn as nn
+from float8_experimental.float8_dynamic_utils import WeightWithDynamicFloat8CastTensor
+from float8_experimental.float8_linear import Float8Linear
+from float8_experimental.float8_linear_utils import linear_requires_sync
+from float8_experimental.float8_utils import EPS
+
+
+def precompute_float8_amax_for_fsdp(module: nn.Module) -> None:
+    """
```
weifengpy, line 14: improve docstring with example API usage

weifengpy:
the linter error comes from trunk. I opened another PR to fix them: https://github.com/pytorch-labs/float8_experimental/pull/313

float8_experimental/float8_dynamic_utils.py
```diff
@@ -130,20 +146,38 @@ def unwrap(t):
         )
 
     def __tensor_flatten__(self):
-        return ["_tensor"], self._mm_config
+        if self._precomputed_amax:
+            return ["_tensor", "_precomputed_amax"], self._mm_config
```
vkuzo, line 150: does having `Optional[torch.Tensor]` as a subclass field work with torch.compile? Or do we not care about torch.compile in this code path?

float8_experimental/fsdp_utils.py
vkuzo: nice! can we add this to the README?

float8_experimental/float8_dynamic_utils.py
weifengpy: torch.compile assumes every tensor from `__tensor_flatten__` is not None. That's why I added `if-else` so when 
`_precomputed_amax is None`, it won't be included in `__tensor_flatten__`

with the `if-else`, torch.compile works. I verified it here and reported perf gaps with eager https://github.com/pytorch/pytorch/issues/129457

float8_experimental/fsdp_utils.py
weifengpy: just added API usage to README

float8_experimental/float8_utils.py
```diff
@@ -34,17 +34,22 @@
...
+    float8_dtype: torch.dtype,
+    orig_dtype: torch.dtype,
+    clamp_amax: bool = True,
 ):
     """Converts the amax value of a tensor to the fp8 scale.
     Args:
         amax: The amax value of the tensor.
         float8_dtype: The float8 dtype.
         orig_dtype: The original dtype of the tensor.
+        clamp_amax: default is True. False for FSDP fp8 all-gather since FSDP applied `torch.clamp` during pre-compute after optimizer.step
```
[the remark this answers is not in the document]
weifengpy: good suggestion! I changed the API to precompute scale and it shows another 9% speed up *in unit test* vs precomputing amax

`fsdp_pre_all_gather` is also greatly simplified because of using `self._precomputed_scale`

README.md
```diff
@@ -51,7 +51,18 @@ model = FSDP(model, use_orig_params=True)
...
 
-# train/finetune (not shown)
+# toy training loop
+for _ in range(N_ITER):
+    optimizer.zero_grad()
+    y = m(x)
+    y.sum().backward()
+    optimizer.step()
+
+    # specific to fsdp2 + float8 with dynamic scaling
```
vkuzo, line 61: should we say that this is specific to FSDP2 with float8 all-gather turned on?  Also, maybe we can show how to turn that on, since I don't think it's documented in the README yet? Can be a follow-up PR.

float8_experimental/fsdp_utils.py
vkuzo: nit: maybe we can make sure `dynamic` is in the name, since this is specific to dynamic scaling?

float8_experimental/fsdp_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    Calculate scale for all float8 parameters after optimizer step
+    It performs a single all-reduce instead of many all-reduces for each parameter
+    Exmaple usage:
+        model(input).sum().backward()
+        optim.step()
+        precompute_float8_scale_for_fsdp(model)
+    """
+    from torch.distributed._tensor import DTensor
+
+    if any(
```
vkuzo, line 24: I think this can be relaxed to something like

```
if isinstance(m, Float8Linear) and m.scaling_type_w is TensorScalingType.DELAYED:
    raise NotImplementedError(...)
```

the scaling of activations and gradient's should not matter for this case

float8_experimental/fsdp_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        isinstance(m, Float8Linear)
+        and linear_requires_sync(
+            m.scaling_type_x, m.scaling_type_w, m.scaling_type_dL_dY
+        )
+        for m in module.modules()
+    ):
+        raise NotImplementedError("Only supports delayed scaling")
+    float8_linears: List[Float8Linear] = [
```
vkuzo, line 32: is this expensive for real models? if yes, maybe we can offer option to precompute this?

float8_experimental/fsdp_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        m
+        for m in module.modules()
+        if isinstance(m, Float8Linear)
+        and isinstance(m.weight, DTensor)
+        and isinstance(m.weight._local_tensor, WeightWithDynamicFloat8CastTensor)
+    ]
+    weights: List[DTensor] = [float8_linear.weight for float8_linear in float8_linears]
+
+    def compute_scales(weights: List[DTensor]):
```
vkuzo, line 41: optional nit: maybe move outside to prevent nested functions?

vkuzo (approved):
looks great! had some final comments. thanks for doing this!

float8_experimental/fsdp_utils.py
```diff
@@ -0,0 +1,61 @@
...
+
+import torch
+import torch.nn as nn
+from float8_experimental.float8_dynamic_utils import WeightWithDynamicFloat8CastTensor
+from float8_experimental.float8_linear import Float8Linear
+from float8_experimental.float8_linear_utils import linear_requires_sync
+from float8_experimental.float8_utils import EPS
+
+
+def precompute_float8_scale_for_fsdp(module: nn.Module) -> None:
```
awgu, line 13: Should we add a `@torch.no_grad()` decorator on this?

float8_experimental/fsdp_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Calculate scale for all float8 parameters after optimizer step
+    It performs a single all-reduce instead of many all-reduces for each parameter
```
awgu, line 16: suggestion:
```suggestion
    Calculate scale dynamically for all float8 parameters.
    This should be run after the optimizer step. It performs a single all-reduce to compute the
    amaxes for all float8 weights.
```

float8_experimental/fsdp_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    Exmaple usage:
```
awgu, line 17: nit (typo):
```suggestion
    Example usage:
```

@vkuzo I assume that there are no docs builds for `float8_experimental`, so this example is for users who will read the code itself?

Otherwise, we might need to check the formatting -- I recall the format for examples being a bit different.

float8_experimental/fsdp_utils.py
awgu: My intuition is that this should be pretty fast as the number of `nn.Module`s in the model is usually at most in the thousands and this is pure Python overhead. @weifengpy you can check the traces you have if you see any noticeable gaps from this.

float8_experimental/fsdp_utils.py
awgu: curious what is the downside of nested functions

@weifengpy By the way, this was originally a nested function just so that we could try to `torch.compile` it effectively in the `scales = compute_scales(weights)` line. Does it still need to be a separate function for `torch.compile` reasons? If so, we should probably add a comment before the `def compute_scales` mentioning that it is separate for `torch.compile`; otherwise, we can consider inlining the function.

float8_experimental/fsdp_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        scales = torch.split(scale_tensor, 1)  # Replicate
+        return scales
+
+    if weights:
+        scales = compute_scales(weights)
+        for scale, float8_linear in zip(scales, float8_linears):
+            float8_linear.weight._local_tensor._precomputed_scale = scale._local_tensor
+    else:
+        warnings.warn(
+            "Calling precompute_float8_weights without any weights using FSDP fp8 all-gather!"
```
awgu, line 60: function name in the warning needs to be updated

I am okay with not including this warning by the way. This was also to help debugging to make sure we actually found `weights`.

float8_experimental/fsdp_utils.py
weifengpy: I will remove nested functions to make the code easy to read. I profiled the unit test and `precompute_float8_scale_for_fsdp` takes 1.9ms. thta's a tiny portion of the overall training loop. No obvious reason to speed up with `torch.compile` yet. I can bring back the nested function in case we need `torch.compile` again.

Screenshot 2024-07-11 at 1 50 46 PM

README.md
weifengpy: > should we say that this is specific to FSDP2 with float8 all-gather turned on?

will change it in this PR

> maybe we can show how to turn that on, since I don't think it's documented in the README yet

good catch. will polish README again after landing changes in torchtitan to turn on/off fp8 all-gather

float8_experimental/fsdp_utils.py
weifengpy: good catch. adding @torch.no_grad()

float8_experimental/fsdp_utils.py
weifengpy: renaming to `precompute_float8_dynamic_scale_for_fsdp`

float8_experimental/fsdp_utils.py
weifengpy: got you. I am removing the warnings for simplicity

float8_experimental/fsdp_utils.py
weifengpy: just checked the profiler traces. it's roughly 0.15ms cpu overhead (5% of `precompute_float8_dynamic_scale_for_fsdp` and is tiny portion of 1 training loop). no cuda are launched

thus I am keeping it as is now for simplicity

PR #271, opened 2024-05-30
delayed scaling: delete Float8LinearMixin

vkuzo:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #273
* #272
* __->__ #271

Summary:

The mixin was originally used to share code with Float8 versions of
RowParallelLinear and ColParallelLinear.  Since we moved those to
DTensor, the mixin is not needed anymore.  Removing it to
simplify the code in preparation of upcoming delayed scaling
improvements.

In addition, making the from_float conversion use meta device
to speed it up.

Test Plan:

```
pytest test/test_base.py -s -x
pytest test/test_compile.py -s -x
```

Reviewers:

Subscribers:

Tasks:

Tags:

Differential Revision: [D58396872](https://our.internmc.facebook.com/intern/diff/D58396872)

PR #272, opened 2024-05-30
delayed scaling: stop syncing weight amax values across ranks

vkuzo:
* #273
* __->__ #272
* #271

Summary:

FSDP already ensures that each rank receives the same weight, so the
amaxes of weights are the same on each rank.

I checked performance before/after on the multi GPU benchmark and
didn't see a significant impact on the toy model, but less comms value is better.

Test Plan:

./test_everything.sh passes

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/benchmarks/bench_multi_gpu.py
+++ b/benchmarks/bench_multi_gpu.py
@@ -25,7 +25,7 @@ from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 torch.manual_seed(0)
 
 # TODO: Add more shapes for the benchmark
-B, M, K, N = 32, 32, 32, 32
+B, M, K, N = 32, 1024, 1024, 1024
 lr = 0.01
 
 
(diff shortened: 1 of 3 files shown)


benchmarks/bench_multi_gpu.py
```diff
@@ -152,8 +152,7 @@ def run_n_iterations(n, fn):
     cleanup()
 
 
-def run():
-    compile = True
+def run(compile: bool):
```
awgu, line 155: How does `compile` get passed into this now?

float8_experimental/float8_linear_utils.py
```diff
@@ -262,10 +262,10 @@ def inner_func():
 
         if dist.is_initialized():
             # Combine all the amax tensors into one tensor and reduce it
+            # Note: do not reduce the weight values, because FSDP already ensures
+            # the weight values on all ranks are the same.
```
awgu, line 266: tiny nit:
```suggestion
            # the weight values on all ranks are the same after all-gather.
```

awgu (approved):
makes sense to me!

benchmarks/bench_multi_gpu.py
vkuzo: it was hardcoded, just making it configurable from command line

vkuzo:
recreated in https://github.com/pytorch-labs/float8_experimental/pull/277 to get around ghstack weirdness

PR #275, opened 2024-06-09
add PrepareFloat8ModuleInput for sequence parallel

wanchaol:
when applying Sequence Parallel to a module with more than 2 linear layers for input proj, we often want to transform from Shard to Replicate once (allgather once) and then reuse the allgathered result, for fp8 we would need to do the casting before the shard -> replicate so that we can perform the fp8 allgather.

This PR subclasses the PrepareModuleInput to add the fp8 casting logic to make sure we run the fp8 allgather instead of bf16 allgather then do the casting for computation.

Also adjust the test cases to test the real ffn case for sequence parallel

torchtitan perf benchmarks (8 H100 devgpu, Llama3 8b, 2-way DP, 4-way TP):

* eager (with no fp8 allgather): 3265 wps
* eager (with fp8 allgather, this PR): 3900 wps
* compile (without fp8 allgather): 5850 wps
* compile (with fp8 allgather): 6592 wps, with 37% MFU on H100

So even in eager we got around 20% perf improvement with every allgather runs in fp8, and compiled fp8 allgather perf is more than doubled (102% more WPS) :)

### DIFF
--- a/float8_experimental/float8_tensor_parallel.py
+++ b/float8_experimental/float8_tensor_parallel.py
@@ -5,7 +5,7 @@ from float8_experimental.float8_dynamic_linear import (
 )
 from torch.distributed._tensor import DTensor
 from torch.distributed.device_mesh import DeviceMesh
-from torch.distributed.tensor.parallel import ColwiseParallel, RowwiseParallel
+from torch.distributed.tensor.parallel import ColwiseParallel, RowwiseParallel, PrepareModuleInput
 
 # subclass the ColwiseParallel and RowwiseParallel classes
 # to add the float8 support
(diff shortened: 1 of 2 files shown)


float8_experimental/float8_tensor_parallel.py
```diff
@@ -109,3 +114,94 @@ def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
...
+                prepared_inputs.append(inp)
+        return tuple(prepared_inputs)
+
+    def _prepare_input_kwarg_fn(self, inputs, kwarg_inputs, device_mesh):
+        prepared_arg_inputs = self._prepare_input_fn(inputs, device_mesh)
+        prepared_kwarg_inputs = {}
+        for kwarg_key in kwarg_inputs.keys():
+            kwarg_val = kwarg_inputs[kwarg_key]
+            input_layout = None
+            if kwarg_key in self.input_kwarg_layouts:
```
vkuzo, line 166: this code seems somewhat similar to L137:L157, is it worth reusing or is it different enough that it isn't?

float8_experimental/float8_tensor_parallel.py
```diff
@@ -109,3 +114,94 @@ def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
             )
 
         return super()._apply(module, device_mesh)
+
+
+class PrepareFloat8ModuleInput(PrepareModuleInput):
```
vkuzo, line 119: nit: maybe we can have e4m3 in the name, and maybe add a TODO to support the AMD version of e4m3 eventually?

float8_experimental/float8_tensor_parallel.py
```diff
@@ -109,3 +114,94 @@ def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
...
+                prepared_kwarg_inputs[kwarg_key] = kwarg_val
+
+        return (prepared_arg_inputs, prepared_kwarg_inputs)
+
+    def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
+        from float8_experimental.float8_dynamic_linear import Float8DynamicLinear
+
+        # search for ScaledMM configs for all the submodules and make sure they are the same
+        fwd_linear_config = None
+        for mod in module.modules():
```
vkuzo, line 196: WDYT something like the following to avoid the logic below?
1. `PrepareFloat8ModuleInput` takes a ScaledMMConfig constructor argument
2. `Float8DynamicLinear` has logic where if the input is already a `Float8Tensor`, there is a check to verify the config matches

?

float8_experimental/float8_tensor_parallel.py
vkuzo: maybe a quick docblock to explain that this is ensuring the float8 cast happens before the all-gather if there are multiple float8 users of the input activation?

float8_experimental/float8_tensor_parallel.py
wanchaol: I thought about this option too, the concern I have on this is that, this would make the API diverges from the TP API offered in core, so making the switch between fp8 and bf16 be harder.

Also I think user would need to know how to construct the ScaledMMConfig, this basically make `ScaledMMConfig` be a public facing API. I wasn't sure this is sth we want or not?

float8_experimental/float8_tensor_parallel.py
wanchaol: > maybe we can have e4m3 in the name, and maybe add a TODO to support the AMD version of e4m3 eventually

I wonder what's your thought on these two choice: 1. make e4m3 appears in the name of this class 2. make this class constructor take an additional argument of fp8 dtype, i.e. `float8_dtype=torch.float8_e4m3fn`, and we default to this e4m3fn dtype, and then later we can add on the AMD version of e4m3 by passing a different float8_dtype` arg?

float8_experimental/float8_tensor_parallel.py
wanchaol: Let me think about how we can refactor this

float8_experimental/float8_tensor_parallel.py
vkuzo: > make this class constructor take an additional argument of fp8 dtype

sgtm

float8_experimental/float8_tensor_parallel.py
vkuzo: > this basically make ScaledMMConfig be a public facing API

Yeah, good point, that's not intended to be a user facing thing.  How about something like requiring a name of the module to get the config from?

I think the user API of the current code is great (no extra args), but the restriction that all configs in the module need the same config is not ideal.  If we are ok with changing that later, current API sgtm.

float8_experimental/float8_tensor_parallel.py
vkuzo: to clarify not asking to refactor if that would make this more complex, I was just wondering.  Keep as is is fine.

float8_experimental/float8_tensor_parallel.py
wanchaol: I think this make sense! Let me draft up the changes for accept the module_fqn to get the scaled mm config from. My current thinking on how we could approach this:
1. We add a `fwd_config_module_fqn` arg to the constructor so that user can specify which module config to take from
2. This arg could be optional, where if user don't pass it in, we still do the search and restrict all configs in this specific module should all be the same.

float8_experimental/float8_tensor_parallel.py
```diff
@@ -109,3 +114,86 @@ def _apply(self, module: nn.Module, device_mesh: DeviceMesh) -> nn.Module:
...
+    def __init__(
+        self,
+        *,
+        input_layouts=None,
+        desired_input_layouts=None,
+        input_kwarg_layouts=None,
+        desired_input_kwarg_layouts=None,
+        use_local_output=False,
+        float8_dtype=torch.float8_e4m3fn,
+        fwd_config_submodule_fqn=None,
```
vkuzo, line 135: nit: maybe we can add a comment on what this is argument is doing?

test/test_dtensor.py
```diff
@@ -192,17 +204,21 @@ def test_fp8_mlp_tensor_parallelism_base(
         sp_model,
         mesh,
         {
-            "in_proj": Float8ColwiseParallel(input_layouts=Shard(0)),
-            "out_proj": Float8RowwiseParallel(
-                output_layouts=Shard(0), use_local_output=False
+            "ffn": PrepareFloat8ModuleInput(
```
vkuzo, line 207: should we also test using `fwd_config_submodule_fqn`?

vkuzo (approved):
awesome!

PR #276, opened 2024-06-11
delayed scaling: delete Float8LinearMixin

vkuzo:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #278
* #277
* __->__ #276

Summary:

The mixin was originally used to share code with Float8 versions of
RowParallelLinear and ColParallelLinear.  Since we moved those to
DTensor, the mixin is not needed anymore.  Removing it to
simplify the code in preparation of upcoming delayed scaling
improvements.

In addition, making the from_float conversion use meta device
to speed it up.

Test Plan:

```
pytest test/test_base.py -s -x
pytest test/test_compile.py -s -x
```

Reviewers:

Subscribers:

Tasks:

Tags:

Differential Revision: [D58396926](https://our.internmc.facebook.com/intern/diff/D58396926)

PR #277, opened 2024-06-11
delayed scaling: stop syncing weight amax values across ranks

vkuzo:
* #278
* __->__ #277
* #276

Summary:

FSDP already ensures that each rank receives the same weight, so the
amaxes of weights are the same on each rank.

I checked performance before/after on the multi GPU benchmark and
didn't see a significant impact on the toy model, but less comms value is better.

Test Plan:

./test_everything.sh passes

Reviewers:

Subscribers:

Tasks:

Tags:

Differential Revision: [D58396925](https://our.internmc.facebook.com/intern/diff/D58396925)

### DIFF
--- a/benchmarks/bench_multi_gpu.py
+++ b/benchmarks/bench_multi_gpu.py
@@ -25,7 +25,7 @@ from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 torch.manual_seed(0)
 
 # TODO: Add more shapes for the benchmark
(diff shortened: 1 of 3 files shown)


vkuzo:
@vkuzo has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D58396925).

drisspg (approved):

PR #283, opened 2024-06-14
Adding Float8 Linear variants supporting inference-only with lower overhead

cyang49:
The changes include two new Float8 Linear implementations that removes some extra wiring in `Float8Linear` unnecessary for inference-only use cases to result in lower latency.

- `Float8SWLinear` supports direct fp8 type direct downcast for activation, and Static per-tensor scale for Weight. Our analysis shows that using this results in no loss of accuracy in Llama models.
- `Float8DASWLinear` supports Dynamic per-tensor scale for Activation, and Static per-tensor scale for Weight. This is used when activation tensor requires dynamic scaling. Compared to `Float8SWLinear`, this has higher overhead introduced by dynamic activation tensor scale calculation. The overhead can be mitigated when used with `torch.compile`.

cc: @ani300

### DIFF
--- a/float8_experimental/float8_linear.py
+++ b/float8_experimental/float8_linear.py
@@ -340,3 +340,87 @@ class Float8Linear(Float8LinearMixin, torch.nn.Linear):
         # I think its okay to send all params and buffers to device
         new_mod.to(mod.weight.device)
         return new_mod
+
+class Float8SWLinear(torch.nn.Linear):
+    """
+    A variation of Float8Linear that operates on torch tensors directly instead of
+    Float8Tensor since the delayed scaling support is not needed. It supports direct fp8
+    type downcast for activation, and Static per-tensor scale for Weight.
+    """
+
+    def __init__(self, in_features, out_features, bias=True, use_triton=False):
+        super(Float8SWLinear, self).__init__(in_features=in_features, out_features=out_features, bias=bias)
+        self.w_inv_s = None
+        self.dtype = torch.float8_e4m3fn
+        self.use_triton = use_triton
+
+    @classmethod
+    def from_float(cls, mod, emulate: bool = False):
+        new_mod = cls(mod.in_features, mod.out_features, bias=mod.bias is not None)
+        assert(not emulate) # no emulation support
+        new_mod.emulate = emulate 
+        
+        w_f8, w_inv_s = new_mod.to_float8(mod.weight)
+
+        new_mod.weight = torch.nn.Parameter(w_f8, requires_grad=False)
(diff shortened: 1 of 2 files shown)


vkuzo:
Hi folks, thanks for the PR and we definitely want this functionality.  We have some changes coming up to set up the inference path, let me give a preview and @drisspg is planning to publish a more detailed RFC related to float8 + inference UX soon.

1. we plan to unify all training logic in `Float8Linear`, `Float8DynamicLinear` would fold into `Float8Linear` and each tensor (act/weight/grad) will be configurable on whether to scale dynamically or delayed
2. we plan to add a separate construct for inference, since the weights are frozen.  It will be something like `Float8InferenceLinear`, although we might also provide a tensor subclass based version.  The behavior of the scaling of the input activation would be configurable at construction time.

In the framing above, this PR would be adding additional options to (2).  Are you ok to wait a little bit for us to put out the official plan?

float8_experimental/float8_linear.py
```diff
@@ -340,3 +340,87 @@ def from_float(cls, mod, emulate: bool = False):
...
+    A variation of Float8Linear that operates on torch tensors directly instead of
+    Float8Tensor since the delayed scaling support is not needed. It supports direct fp8
+    type downcast for activation, and Static per-tensor scale for Weight.
+    """
+
+    def __init__(self, in_features, out_features, bias=True, use_triton=False):
+        super(Float8SWLinear, self).__init__(in_features=in_features, out_features=out_features, bias=bias)
+        self.w_inv_s = None
+        self.dtype = torch.float8_e4m3fn
+        self.use_triton = use_triton
```
vkuzo, line 355: is this used anywhere?

float8_experimental/float8_linear.py
```diff
@@ -340,3 +340,87 @@ def from_float(cls, mod, emulate: bool = False):
...
+        x_scl_sat = (x * scale).clamp(min=finfo.min, max=finfo.max)
+        # Return both float8 data and the inverse scale (as float),
+        # as both required as inputs to torch._scaled_mm
+        return x_scl_sat.to(self.dtype), scale.float().reciprocal() # returns x in self.dtype and scale in f32
+
+    def forward(self, x):
+        x_f8 = x.to(self.dtype)
+        ishape= list(x_f8.shape)
+
+        if ishape[0] == 0: # special case handling for mixtral
```
vkuzo, line 390: is this not supported by `scaled_mm` today? cc @drisspg

cyang49:
no longer needed

PR #284, opened 2024-06-14
Updates with new scaled-mm api

drisspg:
# Summary

This updates the calls to _scaled_mm to the new signature from this PR: https://github.com/pytorch/pytorch/pull/128683

This is needed to unblock inductor work on scaled_mm.

```Shell
❯ ./test/test_everything.sh
	.
	.
	.

test/test_fsdp2/test_fsdp2_eager.py .......                                   [100%]

================================ 7 passed in 27.66s =================================
all tests successful
```

### DIFF
--- a/benchmarks/bench_linear_float8.py
+++ b/benchmarks/bench_linear_float8.py
@@ -14,8 +14,6 @@ import pandas as pd
 
 import torch
 import torch.utils.benchmark as benchmark
-from float8_experimental.float8_dynamic_linear import Float8DynamicLinear
-from float8_experimental.float8_linear import Float8Linear
 from float8_experimental.float8_linear_utils import (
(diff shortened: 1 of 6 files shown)


vkuzo (approved):
LG.  Does this need a PyTorch nightly version check to warn people if their version is out of date, or is there no good way to do that?

drisspg:
We have been pretty fast and loose with versioning.  The only required runtime dependency we have is torch and we dont specify a version, but we do make explicit note of the need for nightly:
https://github.com/pytorch-labs/float8_experimental?tab=readme-ov-file#installation

I could(and probably should) update the dependency we have to specify we require newer versions of pytorch but still not perfect since, nightly doesnt follow semantic versioning

PR #287, opened 2024-06-20
Add a Float8LinearInference module to support static, dynamic, and wo quant

drisspg:
# Summary
# Perf script:
https://gist.github.com/drisspg/f7a553710d64cce013227a2249d582d2

## Performance

In eager this produces:

| Operation                         | Time (μs)  |
|-----------------------------------|------------|
| bf16                              | 2667.9172  |
| fp8_dynamic_activations           | 2494.7294  |
| fp8_static_activations            | 2449.1784  |
| fp8_weight_only_activations       | 4084.7190  |

With compile this produces:
| Operation                    | Time (μs)  |
|------------------------------|------------|
| bf16                         | 2547.1938  |
| fp8_dynamic_activations      | 1542.0729  |
| fp8_static_activations       | 1407.0310  |
| fp8_weight_only_activations  | 2750.6369  |

#### Dynamic activation quantization
``` Python

original_mlp = FeedForward().to("cuda", dtype=dtype)
original_mlp.reset_parameters()

dynamic_fp8_mlp = copy.deepcopy(original_mlp)

quant_config = QuantConfig(ActivationCasting.DYNAMIC)
quantize_to_float8(dynamic_fp8_mlp, quant_config)
```

#### Static activation quantization
```Python
original_mlp = FeedForward().to("cuda", dtype=dtype)
original_mlp.reset_parameters()

static_fp8_mlp = copy.deepcopy(original_mlp)
quant_config = QuantConfig(
    ActivationCasting.STATIC,
    static_quantization_scale=torch.tensor(
        [1.0], device="cuda", dtype=torch.float32
    ),
)
quantize_to_float8(static_fp8_mlp, quant_config)
```

#### Weight Only quantization
``` Python
  original_mlp = FeedForward().to("cuda", dtype=dtype)
  original_mlp.reset_parameters()

  wo_fp8_mlp = copy.deepcopy(original_mlp)
  quant_config = QuantConfig(ActivationCasting.WEIGHT_ONLY)
  quantize_to_float8(wo_fp8_mlp, quant_config)
```

All of these are using Per-Tensor scaling will add in a follow up PR row-wise scaling and likely make this the default.

### DIFF
--- a/float8_experimental/__init__.py
+++ b/float8_experimental/__init__.py
@@ -5,6 +5,11 @@
 # LICENSE file in the root directory of this source tree.
 # Lets define a few top level things here
 from float8_experimental.float8_linear import Float8Linear
-from float8_experimental.float8_tensor import Float8Tensor
+from float8_experimental.float8_tensor import Float8Tensor, ScaledMMConfig
+
+# Needed to load Float8Tensor with weights_only = True
+from torch.serialization import add_safe_globals
+
+add_safe_globals([Float8Tensor, ScaledMMConfig])
 
 __all__ = ["Float8Tensor", "Float8Linear"]
--- a/float8_experimental/float8_dynamic_linear.py
+++ b/float8_experimental/float8_dynamic_linear.py
@@ -73,39 +73,7 @@ class Float8DynamicLinear(torch.nn.Linear):
         return y
 
     @classmethod
-    def create_meta_class(
-        cls, in_features: int, out_features: int
-    ) -> "Float8DynamicLinear":
-        with torch.device("meta"):
-            return cls(in_features=in_features, out_features=out_features, bias=False)
-
-    def set_mm_configs(self, emulate: bool) -> "Float8DynamicLinear":
-        self.forward_config = ScaledMMConfig(
-            emulate, not emulate, pad_inner_dim=config.pad_inner_dim
-        )
-        self.backward_config = ScaledMMConfig(
-            emulate, False, pad_inner_dim=config.pad_inner_dim
-        )
-        return self
-
-    def set_weight_and_bias(
-        self, weight: torch.nn.Parameter, bias: Optional[torch.nn.Parameter]
-    ) -> "Float8DynamicLinear":
-        if config.enable_fsdp_fp8_all_gather:
-            self.weight = nn.Parameter(
-                WeightWithDynamicFloat8CastTensor(weight, self.forward_config)
-            )
-        else:
-            self.weight = weight
-        self.bias = bias
-        return self
-
-    @classmethod
-    def from_float(
-        cls,
-        mod,
-        emulate: bool = False,
-    ) -> "Float8DynamicLinear":
+    def from_float(cls, mod, emulate: bool = False) -> "Float8DynamicLinear":
         """
         Create an nn.Linear with fp8 compute from a regular nn.Linear
 
@@ -113,31 +81,39 @@ class Float8DynamicLinear(torch.nn.Linear):
             mod (torch.nn.Linear): nn.Linear to convert
             emulate (bool): whether to emulate fp8 matmul logic in float32
         """
-        return (
-            cls.create_meta_class(mod.in_features, mod.out_features)
-            .set_mm_configs(emulate)
-            .set_weight_and_bias(mod.weight, mod.bias)
+        with torch.device("meta"):
+            super_kwargs = {
+                "in_features": mod.in_features,
+                "out_features": mod.out_features,
+                "bias": False,
+            }
+            new_mod = cls(**super_kwargs)
+
+        new_mod.forward_config = ScaledMMConfig(
+            emulate=emulate,
+            use_fast_accum=not bool(emulate),
+            fp8_output=False,
+            pad_inner_dim=config.pad_inner_dim,
         )
+        new_mod.backward_config = ScaledMMConfig(
+            emulate=emulate,
+            use_fast_accum=False,
+            fp8_output=False,
+            pad_inner_dim=config.pad_inner_dim,
+        )
+        if config.enable_fsdp_fp8_all_gather:
+            new_mod.weight = nn.Parameter(
+                WeightWithDynamicFloat8CastTensor(mod.weight, new_mod.forward_config)
+            )
+        else:
+            new_mod.weight = mod.weight
+        new_mod.bias = mod.bias
+        return new_mod
 
 
 def cast_to_float8_e4m3fn(
-    inpt_tensor: torch.Tensor,
-    mm_config: ScaledMMConfig,
-    reduce_amax: bool = False,
+    inpt_tensor: torch.Tensor, mm_config: ScaledMMConfig, reduce_amax: bool = False
 ) -> Float8Tensor:
-    """Casts an input tensor to the Float8 (e4m3fn) format for efficient computation.
-
-    Args:
-        inpt_tensor: The input tensor to be cast.
-        mm_config: Configuration settings for the matrix multiplication
-        reduce_amax: Whether to reduce the amax (absolute maximum) among the local distributed group.
-
-    Returns:
-        Float8Tensor: The input tensor cast to Float8 (e4m3fn) format.
-
-    Note:
-        If the input tensor is already in Float8 format, it is returned as is without re-casting.
-    """
     if tensor_already_casted_to_fp8(inpt_tensor):
         return inpt_tensor
     scale = tensor_to_scale(inpt_tensor, e4m3_dtype, reduce_amax)
--- a/float8_experimental/float8_dynamic_linear.py
+++ b/float8_experimental/float8_dynamic_linear.py
@@ -80,8 +80,12 @@ class Float8DynamicLinear(torch.nn.Linear):
             return cls(in_features=in_features, out_features=out_features, bias=False)
 
     def set_mm_configs(self, emulate: bool) -> "Float8DynamicLinear":
-        self.forward_config = ScaledMMConfig(emulate, not emulate, pad_inner_dim=config.pad_inner_dim)
-        self.backward_config = ScaledMMConfig(emulate, False, pad_inner_dim=config.pad_inner_dim)
+        self.forward_config = ScaledMMConfig(
+            emulate, not emulate, pad_inner_dim=config.pad_inner_dim
+        )
+        self.backward_config = ScaledMMConfig(
+            emulate, False, pad_inner_dim=config.pad_inner_dim
+        )
         return self
 
     def set_weight_and_bias(
--- a/float8_experimental/float8_dynamic_linear.py
+++ b/float8_experimental/float8_dynamic_linear.py
@@ -62,8 +62,8 @@ class Float8DynamicLinear(torch.nn.Linear):
     def __init__(self, **super_kwargs):
         super().__init__(**super_kwargs)
 
-    def forward(self, x):
-        x_fp8 = cast_to_float8_e4m3fn(x, self.forward_config)
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
+        x_fp8 = cast_to_float8_e4m3fn(input, self.forward_config)
         if isinstance(self.weight, Float8Tensor):  # cast by FSDP
             w_fp8 = self.weight
         else:
@@ -73,7 +73,35 @@ class Float8DynamicLinear(torch.nn.Linear):
         return y
 
     @classmethod
-    def from_float(cls, mod, emulate: bool = False) -> "Float8DynamicLinear":
+    def create_meta_class(
+        cls, in_features: int, out_features: int
+    ) -> "Float8DynamicLinear":
+        with torch.device("meta"):
+            return cls(in_features=in_features, out_features=out_features, bias=False)
+
+    def set_mm_configs(self, emulate: bool) -> "Float8DynamicLinear":
+        self.forward_config = ScaledMMConfig(emulate, not emulate, pad_inner_dim=config.pad_inner_dim)
+        self.backward_config = ScaledMMConfig(emulate, False, pad_inner_dim=config.pad_inner_dim)
+        return self
+
+    def set_weight_and_bias(
+        self, weight: torch.nn.Parameter, bias: Optional[torch.nn.Parameter]
+    ) -> "Float8DynamicLinear":
+        if config.enable_fsdp_fp8_all_gather:
+            self.weight = nn.Parameter(
+                WeightWithDynamicFloat8CastTensor(weight, self.forward_config)
+            )
+        else:
+            self.weight = weight
+        self.bias = bias
+        return self
+
+    @classmethod
+    def from_float(
+        cls,
+        mod,
+        emulate: bool = False,
+    ) -> "Float8DynamicLinear":
         """
         Create an nn.Linear with fp8 compute from a regular nn.Linear
 
@@ -81,39 +109,31 @@ class Float8DynamicLinear(torch.nn.Linear):
             mod (torch.nn.Linear): nn.Linear to convert
             emulate (bool): whether to emulate fp8 matmul logic in float32
         """
-        with torch.device("meta"):
-            super_kwargs = {
-                "in_features": mod.in_features,
-                "out_features": mod.out_features,
-                "bias": False,
-            }
-            new_mod = cls(**super_kwargs)
-
-        new_mod.forward_config = ScaledMMConfig(
-            emulate=emulate,
-            use_fast_accum=not bool(emulate),
-            fp8_output=False,
-            pad_inner_dim=config.pad_inner_dim,
+        return (
+            cls.create_meta_class(mod.in_features, mod.out_features)
+            .set_mm_configs(emulate)
+            .set_weight_and_bias(mod.weight, mod.bias)
         )
-        new_mod.backward_config = ScaledMMConfig(
-            emulate=emulate,
-            use_fast_accum=False,
-            fp8_output=False,
-            pad_inner_dim=config.pad_inner_dim,
-        )
-        if config.enable_fsdp_fp8_all_gather:
-            new_mod.weight = nn.Parameter(
-                WeightWithDynamicFloat8CastTensor(mod.weight, new_mod.forward_config)
-            )
-        else:
-            new_mod.weight = mod.weight
-        new_mod.bias = mod.bias
-        return new_mod
 
 
 def cast_to_float8_e4m3fn(
-    inpt_tensor: torch.Tensor, mm_config: ScaledMMConfig, reduce_amax: bool = False
+    inpt_tensor: torch.Tensor,
+    mm_config: ScaledMMConfig,
+    reduce_amax: bool = False,
 ) -> Float8Tensor:
+    """Casts an input tensor to the Float8 (e4m3fn) format for efficient computation.
+
+    Args:
+        inpt_tensor: The input tensor to be cast.
+        mm_config: Configuration settings for the matrix multiplication
+        reduce_amax: Whether to reduce the amax (absolute maximum) among the local distributed group.
+
+    Returns:
+        Float8Tensor: The input tensor cast to Float8 (e4m3fn) format.
+
+    Note:
+        If the input tensor is already in Float8 format, it is returned as is without re-casting.
+    """
     if tensor_already_casted_to_fp8(inpt_tensor):
         return inpt_tensor
     scale = tensor_to_scale(inpt_tensor, e4m3_dtype, reduce_amax)
--- a/float8_experimental/float8_ops.py
+++ b/float8_experimental/float8_ops.py
@@ -67,6 +67,26 @@ def float8_split(aten_op, args, kwargs=None):
     return list(out)
 
 
+@implements([aten.copy_.default])
+def copy_fp8(aten_op, args, kwargs=None):
+    self = args[0]
+    src = args[1]
+    assert isinstance(self, Float8Tensor) or isinstance(src, Float8Tensor)
+
+    self_data = self
+    if isinstance(self, Float8Tensor):
+        self_data = self._data
+
+    src_data = src
+    if isinstance(src, Float8Tensor):
+        src_data = src._data
+
+    fp8_out = aten_op(self_data, src_data, *args[2:], **kwargs)
+    if isinstance(self, Float8Tensor):
+        return Float8Tensor(fp8_out, self._scale, self._orig_dtype, self._mm_config)
+    return fp8_out
+
+
 # Errors cant `cat_cuda float8 e4m3fn`
 @implements([aten.cat.default])
 def float8_cat(aten_op, args, kwargs=None):
--- a/float8_experimental/float8_ops.py
+++ b/float8_experimental/float8_ops.py
@@ -67,26 +67,6 @@ def float8_split(aten_op, args, kwargs=None):
     return list(out)
 
 
-@implements([aten.copy_.default])
-def copy_fp8(aten_op, args, kwargs=None):
-    self = args[0]
-    src = args[1]
-    assert isinstance(self, Float8Tensor) or isinstance(src, Float8Tensor)
-
-    self_data = self
-    if isinstance(self, Float8Tensor):
-        self_data = self._data
-
-    src_data = src
-    if isinstance(src, Float8Tensor):
-        src_data = src._data
-
-    fp8_out = aten_op(self_data, src_data, *args[2:], **kwargs)
-    if isinstance(self, Float8Tensor):
-        return Float8Tensor(fp8_out, self._scale, self._orig_dtype, self._mm_config)
-    return fp8_out
-
-
 # Errors cant `cat_cuda float8 e4m3fn`
 @implements([aten.cat.default])
 def float8_cat(aten_op, args, kwargs=None):
--- a/float8_experimental/inference.py
+++ b/float8_experimental/inference.py
@@ -12,6 +12,8 @@ from dataclasses import dataclass
 from enum import auto, Enum
 from typing import List, Optional
 
+import float8_experimental.config as config
+
 import torch
 import torch.nn as nn
 from float8_experimental.float8_linear_utils import swap_linear_layers
@@ -66,8 +68,20 @@ class Float8LinearInference(torch.nn.Linear):
         - FP8 inference with fp8 matmul and static weight casting
     """
 
-    def __init__(self, **super_kwargs):
-        super().__init__(**super_kwargs)
+    def __init__(
+        self,
+        quant_config: QuantConfig,
+        forward_config: ScaledMMConfig,
+        in_features: int,
+        out_features: int,
+        bias: bool = True,
+        device: Optional[torch.device] = None,
+        dtype: Optional[torch.dtype] = None,
+    ) -> None:
+        # Construct the superclass this will create dummy weights and biases
+        super().__init__(in_features, out_features, bias, device, dtype)
+        self.set_quantization_config(quant_config)
+        self.forward_config = forward_config
 
     def forward(self, input: torch.Tensor) -> torch.Tensor:
         if self.activation_casting == ActivationCasting.WEIGHT_ONLY:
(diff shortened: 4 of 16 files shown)


drisspg:
On simple static weight and activation mlp I am seeing a copy_ error
```Python

    swap_linear_with_float8_linear(
            static_fp8_mlp,
            Float8DynamicLinear,
            from_float_kwargs={"static_quantize_weight": True, "activation_scale": torch.tensor([1.0], device="cuda", dtype=torch.float32)},
    )

    print(f"out_static = {static_fp8_mlp(input_tensor)}")
    torch.save(static_fp8_mlp.state_dict(), "/home/drisspg/meta/scripts/fp8/saving/dumm_dict2.pt")

    static_load = torch.load("/home/drisspg/meta/scripts/fp8/saving/dumm_dict2.pt")
    static_fp8_mlp.load_state_dict(static_load)
    print(f"out_static_load = {static_load(input_tensor)}")
    
 ```
 
 ```Shell
 RuntimeError: Error(s) in loading state_dict for FeedForward:
        While copying the parameter named "w1.weight", whose dimensions in the model are torch.Size([14336, 4096]) and whose dimensions in the checkpoint are torch.Size([14336, 4096]), an exception occurred : ('attempting to run aten.copy_.default, this is not supported',).
        While copying the parameter named "w3.weight", whose dimensions in the model are torch.Size([14336, 4096]) and whose dimensions in the checkpoint are torch.Size([14336, 4096]), an exception occurred : ('attempting to run aten.copy_.default, this is not supported',).
        While copying the parameter named "w2.weight", whose dimensions in the model are torch.Size([4096, 14336]) and whose dimensions in the checkpoint are torch.Size([4096, 14336]), an exception occurred : ('attempting to run aten.copy_.default, this is not supported',).
        ```
        
 @ani300 I imagine you were doing some state_dict loading for Float8Tensors?

File: float8_experimental/__init__.py [hunk 1]
```diff
@@ -7,4 +7,9 @@
 from float8_experimental.float8_linear import Float8Linear
 from float8_experimental.float8_tensor import Float8Tensor
 
+# Needed to load Float8Tensor with weights_only = True
+from torch.serialization import add_safe_globals
```
drisspg, line 11: Spoke with @mikaylagawarecki  and named_tuple support can be added to weight_only

float8_experimental/float8_dynamic_linear.py
```diff
@@ -72,14 +76,36 @@ def forward(self, x):
...
+        """Used to perform static_quantization, useful for inferenece where weights are not updated."""
+
+        scale = tensor_to_scale(self.weight, dtype)
+        quantized_weight = to_fp8_no_autograd(
+            self.weight,
+            scale,
+            dtype,
+            self.forward_config,
+        )
+        self.weight = nn.Parameter(quantized_weight)
```
drisspg, line 89: probs decorate in no_grad?

float8_experimental/float8_dynamic_linear.py
```diff
@@ -72,14 +76,36 @@ def forward(self, x):
         y = cast_to_float8_e5m2_bw(y, self.backward_config)
         return y
 
+    def static_quantize_weight(self, dtype: torch.dtype = torch.float8_e4m3fn) -> None:
+        """Used to perform static_quantization, useful for inferenece where weights are not updated."""
```
vkuzo, line 80: usually static quantization refers to quantizing activations with a pre-calibrated zero-point, can we call it something else?  `quantize_weight`? etc

float8_experimental/float8_dynamic_linear.py
vkuzo: I think there is a common assumption that parameters are trainable

float8_experimental/float8_dynamic_linear.py
```diff
@@ -62,8 +62,12 @@ class Float8DynamicLinear(torch.nn.Linear):
     def __init__(self, **super_kwargs):
         super().__init__(**super_kwargs)
 
+        self.activation_scale: Optional[torch.Tensor] = None
```
vkuzo, line 65: we'd probably need serialization for this, either by making it a buffer or adding custom state dict save/load

float8_experimental/float8_dynamic_linear.py
```diff
@@ -72,14 +76,36 @@ def forward(self, x):
...
+        self.weight = nn.Parameter(quantized_weight)
+
     @classmethod
-    def from_float(cls, mod, emulate: bool = False) -> "Float8DynamicLinear":
+    def from_float(
+        cls,
+        mod,
+        emulate: bool = False,
+        static_quantize_weight: bool = False,
+        activation_scale: Optional[torch.Tensor] = None,
```
vkuzo, line 97: we would probably want to provide two options:
1. user specifies the scale they want (for example, for cases where the preceding activation has a bounded range)
2. user specifies how to calculate the scale from calibration data

For (2), the current quantization UXs do this with `Observer` objects (https://fburl.com/code/ady6pz23).  cc @jerryzh168 on if torchao has any different plans to do this in future UXs.

float8_experimental/__init__.py
drisspg: https://github.com/pytorch/pytorch/pull/129251

ani300:
> On simple static weight and activation mlp I am seeing a copy_ error
> 
> ```python
>     swap_linear_with_float8_linear(
>             static_fp8_mlp,
>             Float8DynamicLinear,
>             from_float_kwargs={"static_quantize_weight": True, "activation_scale": torch.tensor([1.0], device="cuda", dtype=torch.float32)},
>     )
> 
>     print(f"out_static = {static_fp8_mlp(input_tensor)}")
>     torch.save(static_fp8_mlp.state_dict(), "/home/drisspg/meta/scripts/fp8/saving/dumm_dict2.pt")
> 
>     static_load = torch.load("/home/drisspg/meta/scripts/fp8/saving/dumm_dict2.pt")
>     static_fp8_mlp.load_state_dict(static_load)
>     print(f"out_static_load = {static_load(input_tensor)}")
>     
> ```
> 
> ```shell
> RuntimeError: Error(s) in loading state_dict for FeedForward:
>        While copying the parameter named "w1.weight", whose dimensions in the model are torch.Size([14336, 4096]) and whose dimensions in the checkpoint are torch.Size([14336, 4096]), an exception occurred : ('attempting to run aten.copy_.default, this is not supported',).
>        While copying the parameter named "w3.weight", whose dimensions in the model are torch.Size([14336, 4096]) and whose dimensions in the checkpoint are torch.Size([14336, 4096]), an exception occurred : ('attempting to run aten.copy_.default, this is not supported',).
>        While copying the parameter named "w2.weight", whose dimensions in the model are torch.Size([4096, 14336]) and whose dimensions in the checkpoint are torch.Size([4096, 14336]), an exception occurred : ('attempting to run aten.copy_.default, this is not supported',).
> ```
> 
> @ani300 I imagine you were doing some state_dict loading for Float8Tensors?

Hey, we were quantizing bf16 weights on the fly from our checkpoints, but I think we'll do something akin to AutoFP8 (https://github.com/neuralmagic/AutoFP8) to handle the FP8 checkpoints and load into Float8Tensors

float8_experimental/float8_dynamic_linear.py
```diff
@@ -72,40 +74,117 @@ def forward(self, x):
         y = cast_to_float8_e5m2_bw(y, self.backward_config)
```
vkuzo, line 74: this line doesn't make sense for inference

float8_experimental/float8_dynamic_linear.py
```diff
@@ -63,7 +63,9 @@ def __init__(self, **super_kwargs):
         super().__init__(**super_kwargs)
 
     def forward(self, x):
-        x_fp8 = cast_to_float8_e4m3fn(x, self.forward_config)
+        x_fp8 = cast_to_float8_e4m3fn(
```
vkuzo, line 66: this looks like static scaling, because the reader has to go inside the `cast_to_float8_e4m3fn` to understand that the scale can actually be calculated if it's not provided

IMO something more readable would be

```
def forward(self, x):
    if is_training:
      ... training logic, unchanged
    else:
      assert is_inference
      custom scaling for activation
      don't call any grad stuff since we are in inference
```

float8_experimental/float8_dynamic_linear.py
vkuzo: to clarify, I know it does nothing, I'm just saying it's a bit confusing and requires the reader to think more to understand what's going on

float8_experimental/inference.py
```diff
@@ -0,0 +1,257 @@
...
+
+class ActivationCasting(Enum):
+    """Types of quantization to perform on the activations
+
+    WEIGHT_ONLY: Only quantize the weight, no activation casting, weight will be dequantized in the forward pass
+    STATIC: Activation is quantized during model initialization with a static scale
+    DYNAMIC: Activation is quantized during forward pass with a dynamic scale calculated from the input activation
+    """
+
+    WEIGHT_ONLY = auto()
```
vkuzo, line 34: nit: maybe this is `NONE`, since "weight-only" is not doing anything to activation?

float8_experimental/inference.py
```diff
@@ -0,0 +1,257 @@
...
+class Float8LinearInference(torch.nn.Linear):
+    """
+    This is a wrapper around torch.nn.Linear that supports FP8 inference
+    Supported forms of infernce:
+        - FP8 inference with fp32 matmul - weight only
+        - FP8 inference with fp8 matmul and dynamic weight casting
+        - FP8 inference with fp8 matmul and static weight casting
+    """
+
+    def __init__(self, **super_kwargs):
```
vkuzo, line 67: why not just add all the config arguments here?  It's a bit confusing that the `__init__` function does not fully initialize the module

float8_experimental/inference.py
```diff
@@ -0,0 +1,257 @@
...
+    Note:
+        If applied to a root-level nn.Linear, the module will not be modified in place
+        and returned instead
+
+    Args:
+        module (nn.Module): The module to modify.
+        quant_config (QuantConfig): Quantization configuration for Float8 conversion.
+        skip_fqn_list (List[str], optional): List of module FQNs to skip during conversion.
+        linear_layer_filter (Callable[[nn.Linear], bool], optional): Function to filter
+            which Linear layers should be converted.
```
vkuzo, line 210: this should be inified (in the training code too), maybe add a comment mentioning that?

float8_experimental/inference.py
```diff
@@ -0,0 +1,257 @@
...
+    if isinstance(module, nn.Linear) and (
+        linear_layer_filter is None or linear_layer_filter(module)
+    ):
+        if len(list(module.children())) > 0:
+            raise AssertionError(
+                f"Does not support a root nn.Linear with children: {module}"
+            )
+        return Float8LinearInference.from_float(module, quant_config)
+
+    # Mark all modules to skip as visited
```
vkuzo, line 228: can this and below be reused with similar logic in training?

float8_experimental/inference.py
```diff
@@ -0,0 +1,257 @@
...
+    STATIC = auto()
+
+
+@dataclass(frozen=True)
+class QuantConfig:
+    """Defines the configuration for the quantization to fp8 of a linear module
+
+    Args:
+        activation_casting: The type of quantization to perform on the activations
+        activation_scale: The scale of the input to this linear module, used for static quantization only
```
vkuzo, line 45: maybe name `static_quantization_scale`?

float8_experimental/inference.py
```diff
@@ -0,0 +1,257 @@
...
+        ), "Weight has already been quantized, cannot quantize again."
+        scale = tensor_to_scale(self.weight, dtype)
+        quantized_weight = to_fp8_no_autograd(
+            self.weight,
+            scale,
+            dtype,
+            self.forward_config,
+        )
+        self.weight = nn.Parameter(quantized_weight)
+        self.weight.requires_grad = False
```
vkuzo, line 104: just curious, what happens if user sets this to true and tries to train?

float8_experimental/inference.py
```diff
@@ -0,0 +1,257 @@
...
+        activation_scale
+        if activation_scale is not None
+        else tensor_to_scale(inpt_tensor, e4m3_dtype, reduce_amax)
+    )
+    return Float8Tensor.to_float8(inpt_tensor, scale, e4m3_dtype, mm_config=mm_config)
+
+
+def quantize_to_float8(
+    module: nn.Module,
+    quant_config: QuantConfig,
```
vkuzo, line 193: can be in separate PR, but long term user needs to be able to specify a separate QuantConfig for each linear if they want to

float8_experimental/inference.py
```diff
@@ -0,0 +1,257 @@
...
+    DYNAMIC: Activation is quantized during forward pass with a dynamic scale calculated from the input activation
+    """
+
+    WEIGHT_ONLY = auto()
+    DYNAMIC = auto()
+    STATIC = auto()
+
+
+@dataclass(frozen=True)
+class QuantConfig:
```
vkuzo, line 40: maybe we can see if there is any alignment possible with torchao on the name or structure of this?  It doesn't have to align now, but ideally the structure is similar so it's easier to unify later.  cc @jerryzh168

float8_experimental/inference.py
drisspg: This can be unified but we then need to match the interface of `from_float` for float8linear and float8Dynamic 
which expects an 'emulate' arg. And since this has extra args also need that kinda gross dict kwarg thing. I thought if we are okay with being not dry at all this is the staightest line. 

That being said let me see if I can do some more re-working

float8_experimental/float8_ops.py
```diff
@@ -63,6 +63,26 @@ def make_float8(data):
     return list(out)
 
 
+@implements([aten.copy_.default])
```
ani300, line 66: I think this is superseded by the final version in my PR that already got merged

drisspg: yup will rebase

float8_experimental/inference.py
drisspg: I agree what you said makes more sense, but I do think that "weight only" or "wo" has become the more common term

vkuzo: oh, I was just pointing out that `weight_only` is about the weight, but the current class name is `ActivationCasting`

float8_experimental/inference.py
```diff
@@ -0,0 +1,228 @@
...
+        )
+        linear = cls(
+            quant_config,
+            forward_config,
+            module.in_features,
+            module.out_features,
+            False,
+            device=torch.device("meta"),
+        )
+        linear.set_weight_and_bias(module.weight, module.bias)
```
vkuzo, line 163: lg, maybe inline since only one callsite and function is super short?

float8_experimental/inference.py
drisspg: tldr is that the float8tensor will get a grad_fn but the grad chain will be broken so its  kind of dangiling. No hard errors

float8_experimental/inference.py
drisspg: Totally agree, I think we wanna unify these names with torchao anyways, is it okay if I leave it as is for this PR?

float8_experimental/inference.py
drisspg: I kinda think it reads nicer tbh, is it okay if we keep as is and can remove in follow up if its distracting?

float8_experimental/float8_dynamic_linear.py
```diff
@@ -62,8 +62,8 @@ class Float8DynamicLinear(torch.nn.Linear):
     def __init__(self, **super_kwargs):
         super().__init__(**super_kwargs)
 
-    def forward(self, x):
-        x_fp8 = cast_to_float8_e4m3fn(x, self.forward_config)
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
```
vkuzo, line 65: nit: the `x`, `w`, `dL_dY` terminology is used throughout this repo, would be good to either stay consistent to that or if we are changing it add a TODO to change it everywhere

float8_experimental/float8_dynamic_linear.py
```diff
@@ -73,47 +73,71 @@ def forward(self, x):
         return y
 
     @classmethod
-    def from_float(cls, mod, emulate: bool = False) -> "Float8DynamicLinear":
+    def create_meta_class(
+        cls, in_features: int, out_features: int
+    ) -> "Float8DynamicLinear":
```
vkuzo, line 78: maybe a comment on why we're adding this?

float8_experimental/float8_dynamic_linear.py
```diff
@@ -73,47 +73,71 @@ def forward(self, x):
...
-            pad_inner_dim=config.pad_inner_dim,
-        )
-        new_mod.backward_config = ScaledMMConfig(
-            emulate=emulate,
-            use_fast_accum=False,
-            fp8_output=False,
-            pad_inner_dim=config.pad_inner_dim,
+        return (
+            cls.create_meta_class(mod.in_features, mod.out_features)
+            .set_mm_configs(emulate)
```
vkuzo, line 118: can this go in the constructor instead?

float8_experimental/inference.py
vkuzo: sgtm, maybe a TODO to fix the name later?

float8_experimental/inference.py
```diff
@@ -0,0 +1,232 @@
...
+    static_quantization_scale: Optional[torch.Tensor] = None
+
+    def __post_init__(self):
+        if self.activation_casting == ActivationCasting.STATIC:
+            assert isinstance(
+                self.static_quantization_scale, torch.Tensor
+            ), "When activation_casting is 'static', activation_scale must be a tensor."
+
+
+class Float8LinearInference(torch.nn.Linear):
```
vkuzo, line 62: wdyt about keeping `Linear` last in the name, having it in the middle is a bit jarring?

float8_experimental/inference.py
```diff
@@ -0,0 +1,232 @@
...
+    """
+    This is a wrapper around torch.nn.Linear that supports FP8 inference
+    Supported forms of infernce:
+        - FP8 inference with fp32 matmul - weight only
```
vkuzo, line 66: fp32 or bf16...maybe "high precision"?

float8_experimental/inference.py
```diff
@@ -0,0 +1,232 @@
...
+        - FP8 inference with fp8 matmul and static weight casting
+    """
+
+    def __init__(
+        self,
+        quant_config: QuantConfig,
+        forward_config: ScaledMMConfig,
+        in_features: int,
+        out_features: int,
+        bias: bool = True,
```
vkuzo, line 77: if we're inheriting from `torch.nn.Linear`, thoughts on adding the aditional arguments last?  I understand the code works but it's extra mental load to think about how the argument order differs, unless it'a s simple "superclass args first, subclass args after"

float8_experimental/inference.py
```diff
@@ -0,0 +1,232 @@
...
+        device: Optional[torch.device] = None,
+        dtype: Optional[torch.dtype] = None,
+    ) -> None:
+        # Construct the superclass this will create dummy weights and biases
+        super().__init__(in_features, out_features, bias, device, dtype)
+        self.set_quantization_config(quant_config)
```
vkuzo, line 83: why not inline, since only one callsite?  I really like looking at the constructor and seeing what all the state is, having a function here forces the reader to jump and go read that function to understand the state.

float8_experimental/inference.py
```diff
@@ -0,0 +1,232 @@
...
+        Note:
+            This function is typically called during inference to quantize the weight once since
+            the weight is not updated during inference.
+
+        """
+        assert not isinstance(
+            self.weight, Float8Tensor
+        ), "Weight has already been quantized, cannot quantize again."
+        scale = tensor_to_scale(self.weight, dtype)
+        quantized_weight = to_fp8_no_autograd(
```
vkuzo, line 115: maybe a TODO for later to fix the name of this function?  Ideally we shouldn't have to think about autograd in inference code.

float8_experimental/inference.py
```diff
@@ -0,0 +1,232 @@
...
+        self,
+        quant_config: QuantConfig,
+    ):
+        # We destructure the quant_config into the individual fields
+        # If an activation config is passed in we want to register that as a buffer
+        self.activation_casting: ActivationCasting = quant_config.activation_casting
+
+        if self.activation_casting == ActivationCasting.STATIC:
+            self.register_buffer(
+                "static_quantization_scale", quant_config.static_quantization_scale
```
vkuzo, line 140: it would be nice to allow the user to pass a static scale (for example, for an activation with bounded range) without having to move it to the right device, can be a TODO for later

float8_experimental/inference.py
```diff
@@ -0,0 +1,232 @@
...
+            module.out_features,
+            False,
+            device=torch.device("meta"),
+        )
+        linear.set_weight_and_bias(module.weight, module.bias)
+        linear.quantize_weight()
+        return linear
+
+
+def cast_to_float8_e4m3fn(
```
vkuzo, line 172: nit: it's a bit confusing to have this function which is the same name as the training `cast_to_float8_e4m3fn` but does a different thing, even if it lives in a different file.  Maybe either split the code into static vs dynamic or make the name clearer?

note that in https://github.com/pytorch-labs/float8_experimental/pull/290/files the training version got renamed to `cast_to_float8_e4m3_dynamic`, I'm not married to that, I just hope whatever names we use are as easy to understand as possible

float8_experimental/inference.py
```diff
@@ -0,0 +1,232 @@
...
+        return inpt_tensor
+    scale = (
+        static_quantization_scale
+        if static_quantization_scale is not None
+        else tensor_to_scale(inpt_tensor, e4m3_dtype, reduce_amax)
+    )
+    return Float8Tensor.to_float8(inpt_tensor, scale, e4m3_dtype, mm_config=mm_config)
+
+
+def quantize_to_float8(
```
vkuzo, line 202: nit: maybe either move to the same utils file as the training swap function, or create inference utils? Just to stay consistent with how the training code is laid out.

vkuzo: also can the names of the training vs inference top level UX be similar?

float8_experimental/float8_dynamic_linear.py
drisspg: I changed from x to input because my linter was complaining that this signature doesn't follow the Nn.Linear subclassing

I can change it back I think we shud todo to switch over the forwards to use the input as a name

test/test_inference_flows.py
```diff
@@ -0,0 +1,246 @@
...
+        compiled_static_fp8_mlp = torch.compile(
+            static_fp8_mlp, backend=compile_backend, fullgraph=True
+        )
+
+        self.base_test_mlp_transform(
+            compiled_original_mlp, compiled_static_fp8_mlp, input_tensor
+        )
+
+
+class TestFP8TrainToFP8LinearInference:
```
vkuzo, line 169: optional: I think we could delete this if you want to lighten the support load, since we already test float8 training is close to high precision, and float8 inference is close to high precision

float8_experimental/inference.py
```diff
@@ -0,0 +1,232 @@
...
+        bias: bool = True,
+        device: Optional[torch.device] = None,
+        dtype: Optional[torch.dtype] = None,
+    ) -> None:
+        # Construct the superclass this will create dummy weights and biases
+        super().__init__(in_features, out_features, bias, device, dtype)
+        self.set_quantization_config(quant_config)
+        self.forward_config = forward_config
+
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
```
vkuzo, line 86: thinking about this some more, there is some confusion I have now on how we want `Float8Tensor` to be related to what gemm is being called

in training code, we seem to be avoiding modifying the `F.linear` snippet, and instead `Float8Tensor.__torch_dispatch__` contains the logic override

here, we are sometimes using `F.linear` and sometimes using `Float8Tensor.__torch_dispatch__`

I can see two more principled solutions:
1. remove the call to `torch._scaled_mm` from `Float8Tensor.__torch_dispatch__` and force the user to call it directly
2. create a new variant of `Float8Tensor` for weight-only quantization

tbh I don't really love those solutions either, but if I had to pick I'd probably want (2), just to remove ambiguity.  Maybe we should brainstorm on how to better separate data representation (float8 tensor = float8 data + scale) versus modeling code (which gemm to call).

vkuzo (approved):
looks great! I have a lot of comments but we can figure those out in future discussions, don't want to block.

float8_experimental/inference.py
vkuzo: or, (3) accept that now the user needs to understand both `Float8InferenceLinear` and `Float8Tensor` to understand which gemm is called (current PR)

float8_experimental/float8_dynamic_linear.py
drisspg: actuallly I am going to change it everywhere

vkuzo: I don't really like `input`, `weight`, `grad_output`, it's a lot of characters and will make the codebase less readable given how they are used.  If we are changing everywhere, I think small # of chars specifically for this is important.  I'd rather override the linter here and not match the `input` name.

drisspg: Hmm as you have already gotten a sense for I am very much in favor of more characters rather than less lol

that being said I think this is an actual issue though:

``` Python
import torch

import torch.nn as nn
import torch.nn.functional as F

from float8_experimental.float8_linear_utils import swap_linear_with_float8_linear


class FeedForward(nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.w1 = nn.Linear(4096, 14336, bias=False)
        self.w3 = nn.Linear(4096, 14336, bias=False)
        self.w2 = nn.Linear(14336, 4096, bias=False)

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        out_1 = self.w1(input=x)
        out_2 = self.w3(input=x)
        out_3 = self.w2(input=out_1 * out_2)
        return out_3


batch_size, num_tokens, embedding_dim = 1, 1, 4096
tst_input = torch.randn(batch_size, num_tokens, embedding_dim)

my_mod = FeedForward()

out_1 = my_mod(tst_input)

swap_linear_with_float8_linear(my_mod, Float8Linear, emulate=True)
my_mod(tst_input)
```

will raise with 
```Python
TypeError: Float8Linear.forward() got an unexpected keyword argument 'input'
```

We dont have control how users call their nn.Linear modules

float8_experimental/float8_dynamic_linear.py
drisspg: since we are removing float8_dynamic_linear I am going to remove the changes here

float8_experimental/float8_dynamic_linear.py
vkuzo: gotcha, that trumps what I said, makes sense

float8_experimental/inference.py
drisspg: The problem with this is that we dont want the two additional args to have default values. By putting two new args  into the __init__ this another example of breaking Liskov principle. However since we dont really expect anyone but "from_float" to call the constructor this is less likely to be hit in practice.

The only other way I can immediately think of to getting around this is to do what was done before, where you create the instance in a partially initialized state and then modify it inplacee.

drisspg: I can do 
```

    def __init__(
        self,
        # FP8 specific arguments
        quant_config: QuantConfig,
        forward_config: ScaledMMConfig,
        # nn.Linear arguments
        in_features: int,
        out_features: int,
        bias: bool = True,
        device: Optional[torch.device] = None,
        dtype: Optional[torch.dtype] = None,
    ) -> None:
    ```
    
    but Imagine you would like that less

float8_experimental/inference.py
drisspg: This is interesting and I was talking about this somewhat with Jerry. 

Currently the dispatch entry for MM only accepts Float8Tensor & Float8Tensor for both its args.
We could update it to accept FP8 + FP8 AND FP8 + Regular. I didnt want to do this since we have multiple systems controlling who casts FP8, fsdp, dtensor, nn_module logc, this feels error prone and we would silently not be casting when we expect. I find comfort in this invariant lol

That being said my initial suggestions is to make a generic "weight_only" subclass wrapper, that will handle the casting up. 

Another way to do achieve something similar is to allow for FP8 + Regular and expand scaled_mm_config to  add constraints. I think the first option is more elegant personally, but I have been pretty embedded in the subclass world so maybe not as easy for others to grok

PR #301, opened 2024-07-03
threading the needle

drisspg:
* __->__ #301

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,8 +10,6 @@ repos:
     -   id: trailing-whitespace
     -   id: check-ast
     -   id: check-merge-conflict
-    -   id: no-commit-to-branch
-        args: ['--branch=main']
     -   id: check-added-large-files
         args: ['--maxkb=500']
     -   id: end-of-file-fixer
--- a/float8_experimental/float8_tensor.py
+++ b/float8_experimental/float8_tensor.py
@@ -4,16 +4,13 @@
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
 from collections import namedtuple
+from enum import auto, Enum
 from typing import Dict, Optional
 
 import torch
 
 import torch.distributed._functional_collectives as funcol
-from float8_experimental.float8_utils import (
-    e4m3_dtype,
-    tensor_to_amax,
-    to_fp8_saturated,
-)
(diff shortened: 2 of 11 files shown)


float8_experimental/float8_tensor.py
```diff
@@ -131,26 +155,35 @@ def forward(
         ctx,
         tensor: torch.Tensor,
         scale: torch.Tensor,
-        float8_dtype=e4m3_dtype,
-        amax_buffer: Optional[torch.Tensor] = None,
-        mm_config: Optional[ScaledMMConfig] = None,
+        float8_dtype: torch.dtype,
+        amax_buffer: Optional[torch.Tensor],
```
drisspg, line 159: I removed the defualt args since this is always called from inner func with defualt args

.pre-commit-config.yaml
```diff
@@ -10,8 +10,6 @@ repos:
     -   id: trailing-whitespace
     -   id: check-ast
     -   id: check-merge-conflict
-    -   id: no-commit-to-branch
```
drisspg, line 13: Fixes ghstack for pre-commit hooks

float8_experimental/float8_tensor.py
```diff
@@ -31,6 +28,20 @@
 )
 
 
+class ScalingStrategy(Enum):
```
vkuzo, line 31: thoughts about using `Granularity`, which is more specific than `Strategy`?

float8_experimental/float8_tensor.py
```diff
@@ -115,7 +133,13 @@ def to_fp8_no_autograd(
             stride=bits_fp8.stride(),
         )
 
-    return Float8Tensor(bits_fp8, x_scale, x.dtype, mm_config=mm_config)
+    return Float8Tensor(
+        bits_fp8,
+        x_scale,
```
vkuzo, line 138: just curious, since we decided to not add `scaling_strategy` to `torch._scaled_mm`, why do we need it here?

float8_experimental/float8_tensor.py
drisspg: Yeah thats a better word, this needed some bikeshedding

float8_experimental/float8_tensor.py
drisspg: We could make this a property of Float8Tensor, e.g. infer from the existing scales... hmm 
Actually I might like this more.. 
We need the enum still since we want modules to specify their granularity

PR #315, opened 2024-07-12
make all 3 gemms in Float8Linear support configurability, not user facing

vkuzo:
* __->__ #315

Summary:

This PR adds some plumbing for how to eventually make all 3 gemms in a linear fwd/bwd configurable:
1. add `LinearMMConfig` to `Float8Tensor` to tie together the three `ScaledMMConfig` objects, one per gemm
2. add `GemmInputRole` to `Float8Tensor` to specify how to pick the right config
3. plumb all of these throughout the codebase

Note that none of this is user facing, and there is no logic change.  Planned follow-ups:
* a future PR will make the per-gemm behavior configurable in a user facing way, which will hook up to the objects introduced in this PR
* a future PR will update the naming from x/w/dL_dY to input/weight/grad_output throughout the codebase

Test Plan:

```
./test/test_everything.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:

Differential Revision: [D59973551](https://our.internmc.facebook.com/intern/diff/D59973551)

### DIFF
--- a/float8_experimental/float8_dynamic_utils.py
+++ b/float8_experimental/float8_dynamic_utils.py
@@ -10,7 +10,8 @@ import torch
 
 from float8_experimental.float8_tensor import (
     Float8Tensor,
-    ScaledMMConfig,
+    GemmInputRole,
+    LinearMMConfig,
     tensor_already_casted_to_fp8,
     to_fp8_no_autograd,
 )
@@ -28,9 +29,9 @@ class NoopFwToFloat8E5M2Bw(torch.autograd.Function):
     def forward(
         ctx,
         tensor,
-        mm_config: ScaledMMConfig,
+        linear_mm_config: LinearMMConfig,
     ):
-        ctx.mm_config = mm_config
+        ctx.linear_mm_config = linear_mm_config
         return tensor
 
     @staticmethod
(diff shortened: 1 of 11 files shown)


float8_experimental/float8_dynamic_utils.py
```diff
@@ -39,21 +40,34 @@ def backward(ctx, gradY):
...
         )
         return fp8_tensor, None
 
 
 def cast_to_float8_e4m3_dynamic(
-    inpt_tensor: torch.Tensor, mm_config: ScaledMMConfig, reduce_amax: bool = False
+    inpt_tensor: torch.Tensor,
+    linear_mm_config: LinearMMConfig,
+    reduce_amax: bool = False,
+    gemm_input_role: GemmInputRole = GemmInputRole.X,
```
drisspg, line 56: maybe no default value for gemm role?

float8_experimental/float8_linear.py
```diff
@@ -192,12 +198,18 @@ def __init__(self, *args, **kwargs):
...
         self.create_buffers()
 
-        # Defines the behavior of the matmul in the forward and backward pass
-        self.forward_config = ScaledMMConfig(
-            emulate, True if not emulate else False, False, config.pad_inner_dim
-        )
-        self.backward_config = ScaledMMConfig(
-            emulate, False, False, config.pad_inner_dim
+        # TODO(future): user level configuration of gemms
+        self.linear_mm_config = LinearMMConfig(
```
drisspg, line 202: [I think this might be another stylistic thing so no need to change]:

I think I would actually make this a func and then super document it. Its not very clear reading this what everything does so I would clearly explain in that func the exact recipe that we choose by default

float8_experimental/float8_ops.py
```diff
@@ -76,6 +84,7 @@ def float8_cat(aten_op, args, kwargs=None):
     scale = chunked_tensors[0]._scale
```
drisspg, line 84: Future PR:
We should code share between unflatten/flatten and here to just splat out the extra metadata that lives on the tensor

float8_experimental/float8_tensor.py
```diff
@@ -30,27 +56,55 @@
...
-    return ScaledMMConfig(
-        emulate=a_mm_config.emulate,
-        use_fast_accum=a_mm_config.use_fast_accum and b_mm_config.use_fast_accum,
-        fp8_output=a_mm_config.fp8_output and b_mm_config.fp8_output,
-        pad_inner_dim=a_mm_config.pad_inner_dim and b_mm_config.pad_inner_dim,
-    )
+# Given a Float8Tensor, the enum below describes the expected role of this
+# tensor in the three gemms present in the fw + bw pass of a Linear layer.
+# This is used to choose the right config for a float8 gemm when the
+# gemm is performed.
```
drisspg, line 76: nit: I would make these comments doc strings since its better for intellisense

float8_experimental/float8_tensor.py
```diff
@@ -30,27 +56,55 @@
...
+        assert (
+            a_linear_mm_config.y == b_linear_mm_config.y
+        ), f"linear_mm_config.y mismatch: {a_linear_mm_config.y} vs {b_linear_mm_config.y}"
+        return a_linear_mm_config.y
+    elif a_role is GemmInputRole.DL_DY and b_role is GemmInputRole.W:
+        assert (
+            a_linear_mm_config.dL_dX == b_linear_mm_config.dL_dX
+        ), f"linear_mm_config.dL_dX mismatch: {a_linear_mm_config.dL_dX} vs {b_linear_mm_config.dL_dX}"
+        return a_linear_mm_config.dL_dX
+    else:
```
drisspg, line 100: nit: I think if you just made 1 more elif branch for the "correct role" and then an else path for the error case would be a little easier to read

float8_experimental/float8_linear.py
vkuzo: this isn't user facing, so we can clean up at any time

float8_experimental/float8_dynamic_utils.py
vkuzo: we can clean up in a separate PR, there is extra complexity because we'd need to change the argument order

drisspg (approved):

vkuzo:
@vkuzo has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D59973551).

PR #316, opened 2024-07-16
[TBD if for land] bring back torch.autograd.Function

vkuzo:
* __->__ #316

Summary:

This approach is more readable as we add additional scaling options.

For now, seeing how many things break in 2024-07 with
torch.autograd.Function + subclasses + compile.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/float8_experimental/float8_linear.py
+++ b/float8_experimental/float8_linear.py
@@ -68,6 +68,101 @@ def _maybe_initialize_amaxes_scales_for_float8_cast(
         )
         scale.copy_(new_scale)
 
+# this code was resurrected from https://github.com/pytorch-labs/float8_experimental/pull/128/files
+# and modified to only support dynamic scaling
+@torch._dynamo.allow_in_graph
+class float8_linear(torch.autograd.Function):
+    """
+    Like F.linear, but with X and W in float8
+    """
+
+    @staticmethod
+    def forward(
+        ctx,
+        x_fp8,
+        w_fp8,
+        emulate: bool,
+        # TODO(this PR): split config into fwd/bwd
+        mm_config: ScaledMMConfig,
+    ):
+        ctx.save_for_backward(x_fp8, w_fp8)
+        ctx.emulate = emulate
+        ctx.mm_config = mm_config
+        # orig_shape = x_fp8._data.shape
+        orig_shape = x_fp8.shape
+        # x_fp8_reshaped = Float8Tensor(
+        #     x_fp8._data.reshape(-1, orig_shape[-1]), x_fp8._scale, x_fp8._orig_dtype, mm_config
+        # )
+        x_fp8_reshaped = x_fp8.reshape(-1, orig_shape[-1])
+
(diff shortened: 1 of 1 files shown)


float8_experimental/float8_linear.py
```diff
@@ -71,6 +71,54 @@ def _maybe_initialize_amaxes_scales_for_float8_cast(
         scale.copy_(new_scale)
 
 
+# this code was resurrected from https://github.com/pytorch-labs/float8_experimental/pull/128/files
```
drisspg, line 74: nit: Does the structure work out to put this in float8 ops?

float8_experimental/float8_linear.py
```diff
@@ -71,6 +71,54 @@ def _maybe_initialize_amaxes_scales_for_float8_cast(
...
+
+    @staticmethod
+    def forward(
+        ctx,
+        x_fp8,
+        w_fp8_t,
+    ):
+        ctx.save_for_backward(x_fp8, w_fp8_t)
+        orig_shape = x_fp8.shape
+        x_fp8_reshaped = x_fp8.reshape(-1, orig_shape[-1])
```
drisspg, line 89: I think a con of this approach is that we dont get to capitalize on the nn.linear decomp ( this worth it). But mabye just a small coment on what this reshape and is doing:
Same thing was needed here: https://github.com/pytorch-labs/float8_experimental/pull/305/files#diff-993c030e4146627dff9686ff1c980699f88ca7f6152a4e809019a4dda645e770R129-R132

maybe make view?

float8_experimental/float8_linear.py
```diff
@@ -71,6 +71,54 @@ def _maybe_initialize_amaxes_scales_for_float8_cast(
...
+        res_bits = torch.mm(x_fp8_reshaped, w_fp8_t)
+        res_bits = res_bits.reshape(*orig_shape[:-1], res_bits.shape[-1])
+        return res_bits
+
+    @staticmethod
+    def backward(ctx, go_fp8):
```
drisspg, line 95: nit: align go_fp8 / other naming to the other PR

drisspg (approved):
Looks good, TBH I think this is a good balance of both subclassing + autograd func

float8_experimental/float8_linear.py
vkuzo: we can do that in separate PRs, since not user facing.  Just keeping things small.

drisspg: I dont know if that changes the size of the PR much but sure thats fine

vkuzo: probably just a style difference on how to sequence the renames, either is ok IMO

float8_experimental/float8_linear.py
vkuzo: in how things look after this PR it would make sense, but might be good to see how the code looks after we add different granularities and the if/else branches on when to convert to lower precision.  Maybe we can revisit then?

PR #321, opened 2024-07-17
add unit tests for FSDP2 + torch.compile(transformer block)

weifengpy:
TorchTitan complains about FSDP2 + float8 + torch.compile(transformer block). 

there is a mismatch in float8 scale so dynamo guards assersion failed `torch._C._dynamo.guards.assert_size_stride(new_inputs[3], (), ())`
* in 1st iteration, we calculate float8 scale through `cast_to_float8_e4m3_dynamic` ([code](https://github.com/pytorch-labs/float8_experimental/blob/main/float8_experimental/fsdp_utils.py#L172)). scale is a scalar tensor, eg `tensor(4674.8633)`
* in 2nd iteration, we calulate float8 scale through `precompute_float8_dynamic_scale`, but scale is NOT a scalar tensor, eg `tensor([[4674.8633]]`
* this PR calls `.squeeze` to make sure scales are always scalar tensors, and dynamo guards assersion always hold true

added unit test so we can catch the isssue at PR time

TODO: add fp8 + torch.compile to CI in torchtitan

### DIFF
--- a/float8_experimental/float8_dynamic_utils.py
+++ b/float8_experimental/float8_dynamic_utils.py
@@ -4,8 +4,6 @@
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Optional, Tuple
-
 import torch
 
 from float8_experimental.float8_tensor import (
--- a/float8_experimental/fsdp_utils.py
+++ b/float8_experimental/fsdp_utils.py
@@ -64,7 +64,9 @@ def precompute_float8_dynamic_scale_for_fsdp(module: nn.Module) -> None:
         scale_tensor = torch.clamp(scale_tensor, max=torch.finfo(torch.float16).max)
     scales = torch.split(scale_tensor, 1)  # Replicate
     for scale, float8_linear in zip(scales, float8_linears):
-        float8_linear.weight._local_tensor._precomputed_scale = scale._local_tensor
+        float8_linear.weight._local_tensor._precomputed_scale = (
+            scale._local_tensor.squeeze()
+        )
 
 
 # FSDP pads its local tensor on dim-0. The subclass should be preserved such
@@ -301,7 +303,7 @@ class WeightWithDelayedFloat8CastTensor(torch.Tensor):
             ],
             {
                 "mm_config": self._mm_config,
-                "is_amax_initialized": is_amax_initialized,
+                "is_amax_initialized": self.is_amax_initialized,
             },
         )
 
--- a/test/test_fsdp2/test_fsdp2.py
+++ b/test/test_fsdp2/test_fsdp2.py
@@ -90,12 +90,6 @@ class TestFloat8MultiProcess(FSDPTest, TestFloat8Common):
                     TensorScalingType.DELAYED,
                 ],
                 "compile_transformer_block": [False, True],
-                # "enable_fsdp_fp8_all_gather": [True],
-                # "precompute": [True],
-                # "scaling_type_w": [
-                #     TensorScalingType.DYNAMIC,
-                # ],
-                # "compile_transformer_block": [True],
             },
             self._test_transformer_parity,
         )
--- a/test/test_fsdp2/test_fsdp2.py
+++ b/test/test_fsdp2/test_fsdp2.py
@@ -89,6 +89,13 @@ class TestFloat8MultiProcess(FSDPTest, TestFloat8Common):
                     TensorScalingType.DYNAMIC,
                     TensorScalingType.DELAYED,
                 ],
+                "compile_transformer_block": [False, True],
+                # "enable_fsdp_fp8_all_gather": [True],
+                # "precompute": [True],
+                # "scaling_type_w": [
+                #     TensorScalingType.DYNAMIC,
+                # ],
+                # "compile_transformer_block": [True],
             },
             self._test_transformer_parity,
         )
@@ -98,6 +105,7 @@ class TestFloat8MultiProcess(FSDPTest, TestFloat8Common):
         enable_fsdp_fp8_all_gather: bool,
         precompute: bool,
         scaling_type_w: TensorScalingType,
+        compile_transformer_block: bool,
     ):
         if not enable_fsdp_fp8_all_gather and precompute:
             return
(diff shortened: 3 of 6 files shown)


float8_experimental/fsdp_utils.py
```diff
@@ -301,7 +303,7 @@ def __tensor_flatten__(self):
             ],
             {
                 "mm_config": self._mm_config,
-                "is_amax_initialized": is_amax_initialized,
```
weifengpy, line 304: `pre-commit run --all-files` complains about undefined `is_amax_initialized` in trunk. fixing it so I can commit without bypassing linter

test/test_fsdp2/test_fsdp2_common.py
```diff
@@ -46,7 +47,10 @@ def check_parity_no_mp(
             ):
                 precompute_float8_dynamic_scale_for_fsdp(model)
 
-        test_cls.assertEqual(losses[0], losses[1])
+        if compile_transformer_block:
+            torch.testing.assert_close(losses[0], losses[1], atol=9.5e-2, rtol=9.5e-2)
```
weifengpy, line 51: borrowed 9.5e-2 from `test_compile.py` https://github.com/pytorch-labs/float8_experimental/blob/main/test/test_compile.py#L62

float8_experimental/fsdp_utils.py
```diff
@@ -64,7 +64,9 @@ def precompute_float8_dynamic_scale_for_fsdp(module: nn.Module) -> None:
         scale_tensor = torch.clamp(scale_tensor, max=torch.finfo(torch.float16).max)
     scales = torch.split(scale_tensor, 1)  # Replicate
     for scale, float8_linear in zip(scales, float8_linears):
-        float8_linear.weight._local_tensor._precomputed_scale = scale._local_tensor
+        float8_linear.weight._local_tensor._precomputed_scale = (
+            scale._local_tensor.squeeze()
```
weifengpy, line 68: make sure tensor is like `tensor(4674.8633)` instead of `tensor([[4674.8633]]`

test/test_fsdp2/test_fsdp2_common.py
awgu: This seems kind of high 🤔  I wonder how this value was determined. Can we instead compare the ref as also compiling each transformer block (but without FSDP applied)?

float8_experimental/float8_dynamic_utils.py
```diff
@@ -4,8 +4,6 @@
 # This source code is licensed under the BSD 3-Clause license found in the
 # LICENSE file in the root directory of this source tree.
 
-from typing import Any, Optional, Tuple
```
weifengpy, line 7: fix linter from the trunk

float8_experimental/fsdp_utils.py
awgu: Can we check the traces? I want to make sure there is no CPU sync point introduced from making this tensor a scalar tensor.

vkuzo: do we know the reasoning for why the current behavior is not supported with compile?  This might not scale long term as we add other scaling granularities like rowwise or blockwise.

weifengpy: > Can we check the traces? I want to make sure there is no CPU sync point introduced from making this tensor a scalar tensor

sure. I checked the trace and it's seems to be purely executed on cpu, no kernels launch, and no `cudaStreamSynchronize` if that's what you refer as "CPU sync point"
Screenshot 2024-07-17 at 2 26 24 PM

awgu: I mean when the scalar is used later downstream.

weifengpy: > do we know the reasoning for why the current behavior is not supported with compile? This might not scale long term as we add other scaling granularities like rowwise or blockwise.

TL;DR: this is more like a bug when I implement `precompute_float8_dynamic_scale_for_fsdp`

for the 1st iteration, `self._precomputed_scale` is `None` and thus we calcuclate scale through `cast_to_float8_e4m3_dynamic`, where scale are in `tensor(4674.8633)`

for the 2nd iteration after `precompute_float8_dynamic_scale_for_fsdp`, we have `self._precomputed_scale=tensor([[4674.8633]])` because I only called `torch.split(scale_tensor, 1)` without `.squeeze`. It should be my responsibility to keep `precompute_float8_dynamic_scale_for_fsdp` consistent with `cast_to_float8_e4m3_dynamic`

weifengpy: > I mean when the scalar is used later downstream.

ah, I see. I should be looking for `cudaStreamSynchronize`, right ?

test/test_fsdp2/test_fsdp2_common.py
weifengpy: will try to switch the ref model to Float8Linear + torch.compiled

float8_experimental/fsdp_utils.py
awgu: I think it would be like `cudaDeviceSynchronize` if I understand correctly (but basically you would see the CPU thread blocked).

weifengpy: > I mean when the scalar is used later downstream.

`_precomputed_scale` will be used inside `fsdp_pre_all_gather` when calling

```
float8_tensor = Float8Tensor.to_float8(
    self._tensor,
    self._precomputed_scale,
    torch.float8_e4m3fn,
    mm_config=self._mm_config,
)
```

I annotated the function with `record_function("Float8Tensor.to_float8")`. Here are the snapshots for cpu thread and cuda stream

Screenshot 2024-07-17 at 4 33 53 PM

Screenshot 2024-07-17 at 4 34 31 PM

in both cases, I do not see `cudaStreamSynchronize` and cuda stream stays ahead of cpu thread

any worries ?

awgu: sounds good! should be fine

test/test_fsdp2/test_fsdp2_common.py
weifengpy: after applying torch.compile to ref_model, we can achieve `atol/rtol=1e-4`. I can dig more as follow ups if we want to reach higher numeric parity like `1e-5`

test/test_fsdp2/test_fsdp2.py
```diff
@@ -112,11 +114,17 @@ def _test_transformer_parity(
...
+                transformer_block = torch.compile(transformer_block, dynamic=False)
+                ref_module.layers.register_module(layer_id, transformer_block)
         with set_enable_fsdp_fp8_all_gather(enable_fsdp_fp8_all_gather):
             swap_linear_with_float8_linear(module, scaling_type_w=scaling_type_w)
-        for submodule in module.modules():
-            if isinstance(submodule, TransformerBlock):
-                fully_shard(submodule)
+        for layer_id, transformer_block in module.layers.named_children():
+            if compile_transformer_block:
+                transformer_block = torch.compile(transformer_block, dynamic=False)
```
vkuzo, line 125: is compiling the transformer block instead of the entire model related to this issue, or are we just trying to match torchtitan behavior?

optionally, if possible, would be good to compile the whole model here instead as long as that can catch the issues relevant to us and keep the more advanced "how to apply compile" logic localized to torchtitan.

vkuzo (approved):
looks great, thanks for fixing this!

bdhirsh (commented):
nice catch!

test/test_fsdp2/test_fsdp2.py
weifengpy: > is compiling the transformer block instead of the entire model related to this issue, or are we just trying to match torchtitan behavior?

This is just trying to match torchtitan's behavior. The `.squeeze` is needed regardless of compiling transformer blocks or compiling whole model. 

> optionally, if possible, would be good to compile the whole model here instead as long as that can catch the issues relevant to us 

I want at PR time that float8_experimental changes are compatiable with torchtitan (thus compiling transformer block)

for float8_experimental, I am with you it's good to also cover compiling full model. 
For FSDP2, it should work. For FSDP+TP, I remember there is some problem in to compile full model. Will see if I can follow up

PR #322, opened 2024-07-19
[bc breaking] unify filtering functions

vkuzo:
* #323
* __->__ #322

Summary:

Before this PR, we had two top level filtering affordances on `swap_linear_with_float8_linear`: `skip_fqn_list` and `linear_layer_filter`:

```
def swap_linear_with_float8_linear(
    ...,
    skip_fqn_list: Optional[List[str]] = None,
    linear_layer_filter: Optional[Callable[[nn.Linear], bool]] = None,
)
```

This PR unifies them into a single filtering method which is aware of both the FQN as well as the module instance.  This should be more future proof and allow for more fine grained filtering.

```
def swap_linear_with_float8_linear(
    ...,
    layer_filter_fn: Optional[Callable[[str, nn.Module], bool]] = None,
)

```

Note that the `filter_out_small_unaligned_layers` function is removed from the public API, and users are encouraged to write their own, again to make this more future proof.  I'm not opposed to adding good utility functions back at some point, but IMO we should finalize the main UX first.

Note that in the future (as outlined in the Meta-only UX discussion doc), we may change this function from being just a filter to also providing the float8 per-module config.  I'm saving that for a separate PR, which will also be BC breaking (but we don't have BC yet).

Test Plan:

```
./test/test_everything.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:

Differential Revision: [D60072597](https://our.internmc.facebook.com/intern/diff/D60072597)

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,19 @@ from float8_experimental.fsdp_utils import precompute_float8_dynamic_scale_for_f
 # create model
 m = Model(...)
 
+# optional: filter layers from being eligible for float8 conversion
+def layer_filter_fn(fqn: str, mod: torch.nn.Module):
+    # don't convert the output layer
+    if fqn == "output":
+        return False
+    # don't convert linear layers with weight dimensions not divisible by 16
+    if isinstance(mod, torch.nn.Linear):
+        if mod.in_features % 16 != 0 or mod.out_features % 16 != 0:
+            return False
+    return True
+
 # convert all `torch.nn.Linear` modules to `Float8Linear`
-swap_linear_with_float8_linear(m)
+swap_linear_with_float8_linear(m, layer_filter_fn=layer_filter_fn)
 
 # optional: use FSDP
 model = FSDP(model, use_orig_params=True)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -43,8 +43,19 @@ from float8_experimental.fsdp_utils import precompute_float8_dynamic_scale_for_f
 # create model
 m = Model(...)
 
+# optional: filter layers from being eligible for float8 conversion
+def layer_filter_fn(fqn: str, mod: torch.nn.Module):
```
awgu, line 47: nit: how do you feel about using "module" instead of "layer", which may be more canonical in PyTorch?

vkuzo: sgtm

awgu (approved):
seems good to me!

y-sq (approved):

vkuzo:
@vkuzo has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D60072597).

PR #323, opened 2024-07-22
[bc breaking] change x, w, dL_dY variable names to input, weight, grad_output

vkuzo:
* __->__ #323
* #322

Summary:

The following naming scheme matches the rest of PyTorch better:

```Python
// forward
output = input @ weight_t
// backward
grad_input = grad_output @ weight
grad_weight = input_t @ grad_output
```

This PR changes all the previous references to `x`, `w`, `dL_dY` to
match the naming scheme above.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

Differential Revision: [D60072596](https://our.internmc.facebook.com/intern/diff/D60072596)

### DIFF
--- a/.github/workflows/ufmt.yml
+++ b/.github/workflows/ufmt.yml
@@ -23,4 +23,7 @@ jobs:
         pip install black==23.3.0 usort==1.0.6 ufmt==2.1.0 libcst==1.0.1
     - name: Analyzing the code with ufmt
       run: |
+        ufmt format .
+        git diff
+        git restore .
         ufmt check .
--- a/float8_experimental/float8_linear.py
+++ b/float8_experimental/float8_linear.py
@@ -99,9 +99,11 @@ class NoopFwToFloat8E5M2Bw(torch.autograd.Function):
 
     @staticmethod
     def backward(ctx, go):
-        fp8_amax_grad_output, fp8_amax_history_grad_output, fp8_scale_grad_output = (
-            ctx.saved_tensors
-        )
+        (
+            fp8_amax_grad_output,
+            fp8_amax_history_grad_output,
+            fp8_scale_grad_output,
+        ) = ctx.saved_tensors
         scale_fn_name = ctx.scale_fn_name
         is_amax_initialized = ctx.is_amax_initialized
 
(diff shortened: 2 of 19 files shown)


.github/workflows/ufmt.yml
```diff
@@ -23,4 +23,7 @@ jobs:
         pip install black==23.3.0 usort==1.0.6 ufmt==2.1.0 libcst==1.0.1
     - name: Analyzing the code with ufmt
       run: |
+        ufmt format .
```
vkuzo, line 26: L26:28 is for easier debugging of differences between local machine and CI ufmt

drisspg: is the ufmt config different than CI?

float8_experimental/float8_linear.py
```diff
@@ -82,41 +82,47 @@ class NoopFwToFloat8E5M2Bw(torch.autograd.Function):
...
+        ctx.save_for_backward(
+            fp8_amax_grad_output, fp8_amax_history_grad_output, fp8_scale_grad_output
+        )
         ctx.scale_fn_name = scale_fn_name
         ctx.is_amax_initialized = is_amax_initialized
         ctx.linear_mm_config = linear_mm_config
         return tensor
 
     @staticmethod
     def backward(ctx, go):
```
drisspg, line 101: should we update go? This one has confused me in the past

.github/workflows/ufmt.yml
vkuzo: I've noticed this on a couple of PRs in the past.  A better fix would be to align ufmt versions + env, but regardless this is useful for debugging.

drisspg (approved):
Awesome, thanks!

float8_experimental/float8_linear.py
vkuzo: this PR only contains the user-facing part since that's the most crucial.  LOC is already high, so IMO better to rename the non-user facing things in subsequent PRs, to keep things smaller and more reviewable.

vkuzo:
@vkuzo has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D60072596).

PR #324, opened 2024-07-23
[bc breaking] move most previously-global configs to Float8LinearConfig

vkuzo:
* #325
* __->__ #324

Summary:

Adds a `Float8LinearConfig` to unify the user facing per-linear
configuration, and moves most of the previously global config options
there.

In future PRs (to keep PRs small), we will move emulation, scaling
and gemm configurations to also live here.

Test Plan:

```
./test/test_everything.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:

Differential Revision: [D60176981](https://our.internmc.facebook.com/intern/diff/D60176981)

### DIFF
--- a/float8_experimental/float8_linear.py
+++ b/float8_experimental/float8_linear.py
@@ -12,9 +12,9 @@ import enum
 
 from typing import Optional
 
-import float8_experimental.config as config
-
 import torch
+
+from float8_experimental.config import Float8LinearConfig
 
 from float8_experimental.float8_dynamic_utils import (
     cast_to_float8_e4m3_dynamic,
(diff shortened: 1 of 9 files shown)


float8_experimental/float8_linear.py
```diff
@@ -450,14 +460,18 @@ def from_float(
...
         """
         Create an nn.Linear with fp8 compute from a regular nn.Linear
 
         Args:
             mod (torch.nn.Linear): nn.Linear to convert
             emulate (bool): whether to emulate fp8 matmul logic in float32
+            config (Optional[Float8LinearConfig]): configuration for conversion to float8
         """
+        if config is None:
+            config = Float8LinearConfig()
```
weifengpy, line 474: is non-None config the key to avoid torch.compile error?

weifengpy (approved):

float8_experimental/float8_linear.py
vkuzo: I'm just taking a shortcut:
1. long term config should be required
2. short term, making config required would need us to move it up to positional args, which would then require us to modify all the callsites
3. long term, we will also move `emulate` and `scaling_type...` to `config`

If we make the argument required after we do 3, it should be less work overall.

vkuzo:
@vkuzo has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D60176981).

PR #326, opened 2024-07-24
[DISCUSSION] fix float8 all-gather in FSDP2 + TP: DTensor(WeightWithDynamicFloat8CastTensor)

weifengpy:
draft this PR for discussion, before having something landable

we see 2 problems in float8 all-gather FSDP2 + TP
* FSDP2 all-gather is in bf16, but expect float8
* TP all-reduce amax for `weight`, but expect all-reduce only for `input`

crux is how we dispatch `torch.chunk`, which is called from `distribute_tensor` for TP init
* without this PR, `torch.chunk` returns `Tensor`. FSDP2 happens after TP, thus only see `Float8Linear(weight=DTensor(_local_tensor=Tensor))`
* with this PR, `torch.chunk` returns `WeightWithDynamicFloat8CastTensor`

profiler trace without this PR: AR (all-reduce) for input -> AG (all-gather) -> 4 ARs for wq,k,v,o -> 1 AR for input. 4 ARs for wq,k,v,o should not happen if we precompute amax/scales for `model.parameters()` after `opt.step()`
 
Screenshot 2024-07-24 at 12 55 56 AM

### DIFF
--- a/float8_experimental/fsdp_utils.py
+++ b/float8_experimental/fsdp_utils.py
@@ -14,8 +14,8 @@ from float8_experimental.float8_dynamic_utils import cast_to_float8_e4m3_dynamic
 
 from float8_experimental.float8_tensor import (
     Float8Tensor,
-    merge_mm_configs,
-    ScaledMMConfig,
+    GemmInputRole,
+    LinearMMConfig,
 )
 
 from float8_experimental.float8_utils import e4m3_dtype, EPS
@@ -89,7 +89,7 @@ class WeightWithDynamicFloat8CastTensor(torch.Tensor):
     def __new__(
         cls,
         tensor: torch.Tensor,
-        mm_config: ScaledMMConfig,
+        linear_mm_config: LinearMMConfig,
         precomputed_scale: Optional[torch.Tensor] = None,
     ):
         return torch.Tensor._make_wrapper_subclass(
@@ -108,11 +108,11 @@ class WeightWithDynamicFloat8CastTensor(torch.Tensor):
     def __init__(
         self,
         tensor: torch.Tensor,
-        mm_config: ScaledMMConfig,
+        linear_mm_config: LinearMMConfig,
         precomputed_scale: Optional[torch.Tensor] = None,
     ):
         self._tensor = tensor
-        self._mm_config = mm_config
+        self._linear_mm_config = linear_mm_config
         # for dynamic scaling
         # `precompute_float8_dynamic_scale_for_fsdp` calculates scales
         # for all float8 parameters after optimizer step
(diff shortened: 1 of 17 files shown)


test/test_fsdp2/test_fsdp2.py
```diff
@@ -516,5 +522,50 @@ def test_delayed_scaling_inplace_update(self):
...
+            swap_linear_with_float8_linear(module, scaling_type_w=scaling_type_w)
+
+        # "attention.wq": Float8ColwiseParallel
+        colwise_param = distribute_tensor(
+            module.layers[0].attention.wq.weight, tp_mesh, [Shard(0)]
+        )
+        self.assertTrue(
+            isinstance(colwise_param, DTensor)
+            and isinstance(
+                colwise_param._local_tensor, WeightWithDynamicFloat8CastTensor
```
weifengpy, line 555: without this PR, `torch.chunk` returns `DTensor(_local_tensor=bf16_tensor)`. FSDP2 happens after TP, thus only see `Float8Linear(weight=DTensor(_local_tensor))`
with this PR, `torch.chunk` returns `DTensor(_local_tensor=WeightWithDynamicFloat8CastTensor)`

float8_experimental/fsdp_utils.py
```diff
@@ -81,6 +81,8 @@ def precompute_float8_dynamic_scale_for_fsdp(module: nn.Module) -> None:
     torch.ops.aten.as_strided.default,
     torch.ops.aten._to_copy.default,
     torch.ops.aten._pin_memory.default,
+    torch.ops.aten.split.Tensor,
```
weifengpy, line 84: `aten.split` is from `torch.chunk`, when calling from `distribute_tensor` during TP init

@awgu curious if you still remember the reason to return bf16 from torch.chunk. Is it for padding? any concerns if I prefer `torch.chunk` to returning `WeightWithDynamicFloat8CastTensor` ?

float8_experimental/fsdp_utils.py
```diff
@@ -188,12 +190,22 @@ def fsdp_post_all_gather(
...
         if out is not None:
-            assert isinstance(out, Float8Tensor), f"{type(out)}"
-            out._scale = scale
-            return
+            if isinstance(out, Float8Tensor):
+                out._scale = scale
+            elif isinstance(out, DTensor) and isinstance(
+                out._local_tensor, Float8Tensor
+            ):
+                out._local_tensor._scale = scale
```
weifengpy, line 203: not sure about this change yet. just want to have someting sketchy to discuss first

float8_experimental/fsdp_utils.py
awgu: > @awgu curious if you still remember the reason to return bf16 from torch.chunk.

I thought that dtype and whether is `WeightWithDynamicFloat8CastTensor` are orthogonal. Do you mean the latter (whether is `WeightWithDynamicFloat8CastTensor` or not?

I think originally I only added the ops that I saw I needed. Adding `aten.split` and `aten.clone` seems okay to me.

test/test_fsdp2/test_fsdp2.py
awgu: Can you explain where the bf16 came from?

float8_experimental/fsdp_utils.py
weifengpy: > whether is WeightWithDynamicFloat8CastTensor or not

exactly, `WeightWithDynamicFloat8CastTensor` or not is the key. I edited my previous comments to say right now torch.chunk returns `Tensor`

> I think originally I only added the ops that I saw I needed

changing `torch.chunk` affects both TP and FSDP2. will double check FSDP2 after the change

test/test_fsdp2/test_fsdp2.py
weifengpy: correct my word to be accurate: without this PR, `torch.chunk` returns plain `Tensor` (can be fp32 or bf16) instead of `WeightWithDynamicFloat8CastTensor`

PR #342, opened 2024-07-26
[1/x] clean up casting functions

vkuzo:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #343
* __->__ #342
* #341

Summary:

This is a start of a cleanup of private casting functions in preparation for
rowwise scaling.  In this PR:
1. create `float8_scaling_utils.py` to unify functions which take a
   high precision tensor and return a float8 tensor, taking care of
   scaling
2. delete `Float8Tensor.to_float8` and move callsites to
   `ToFloat8ConstrFunc`, since the two functions do the same thing

The end result is a slightly cleaner state, future PRs will do more
cleanups.

Test Plan:

```
./test/test_everything.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:

PR #343, opened 2024-07-26
[2/x] clean up casting functions: delayed scaling

vkuzo:
* __->__ #343
* #342
* #341

Summary:

Removes delayed scaling from `float8_tensor.py`. After this PR, the
invariant is that everything in `float8_tensor.py` requires the scale to
be calculated elsewhere. This moves the codebase towards separation of
concerns for calculating the scale (via various scaling strategies),
separated from creating an instance of `Float8Tensor`.

Note that stateful delayed scaling is the reason we need this separation.

Test Plan:

```
./test/test_everything.sh
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/benchmarks/bench_padding.py
+++ b/benchmarks/bench_padding.py
@@ -62,7 +62,6 @@ def do_fp8_matmul(A, B, fp8_dtype, out_dtype):
         A,
         scale_a,
         fp8_dtype,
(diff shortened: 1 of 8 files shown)


vkuzo:
started a new PR due to ghstack error

